PR #12, opened 2020-02-06
Onboard additions

tonywu315:
Fixed some typos, clarified instructions, and added instructions for running both servers using the script for Windows.
@ahsu1230 @mzhang2021 @MiniViolin @mythicat554


orion/README.md
```diff
@@ -194,26 +195,29 @@ Great job getting here! You've built 2 local webservers to host the web applicat
...
 *Mac users only*
 ```
 ./run_admin_local.sh
 ```
 *Windows Users only*
+
+Open Windows PowerShell and run the command above. Alternatively, on Command Prompt you can use
 ```
-coming soon...
+run_admin_local.sh
```
ahsu1230: hmm I'm not sure about this. Windows PowerShell is only default for Windows 10. And normally, you can't run .sh files on Windows. I'd say keep the note about Windows PowerShell above. But, if viewers don't have Windows 10, maybe I'll create a .bat or .cmd file for them to run on Command Prompt.

tonywu315: I think it works for me because it opens a git bash application. I added a batch file that works for Command Prompt.

orion/run_admin_local.bat
```diff
@@ -0,0 +1,17 @@
...
+:: So you may interact with the website and api server end to end.
+
+:: First step is to make sure mysql is installed on your computer and running
+:: net start MySQL
+:: net stop MySQL
+:: mysql db --user=root --password
+
+:: Second step is to build the website locally
+cd sites\admin
+call npm run build-local
```
tonywu315: You need a "call" in front because of an issue between batch and npm
https://github.com/npm/npm/issues/2938

orion/README.md
ahsu1230: Now that the batch file is in, could you please change this to `run_admin_local.bat`?

ahsu1230: ah nvm you did.

ahsu1230:
You should be able to merge even though the "check" failed. Lmk if it's still blocking you.

PR #13, opened 2020-02-08
Valid Program Name Check

tonywu315:
I am not really sure how gin works, so I do not know how to test the code.
@ahsu1230  
@mzhang2021 
@MiniViolin  
@mythicat554


ahsu1230:
Thank you for pulling this into a separate PR. Give me a moment to take a look.

orion/controllers/programs/programs_test.go
```diff
@@ -12,8 +12,9 @@ func TestExample(t *testing.T) {
 }
 
 func TestCheckProgram(t *testing.T) {
-  isValid := CheckValidProgram()
-  if (isValid == false) {
-    t.Errorf("Check was incorrect, got: %t, expected: %t.", isValid, false)
+  // TODO: add a program to check
+  isValid := CheckValidProgram(program)
```
ahsu1230: this will probably error because `program` is not initialized, right?
Create a new Program{} and feel free to give it fake values to test if `CheckValidProgram()` works as intended.
You can do
```
program := new Program{name: "asdf", grade1: 1, grade2: 2, ... }
```

orion/controllers/programs/programsController.go
```diff
@@ -1,8 +1,9 @@
 package programs
 
 import (
-  "net/http"
   "github.com/gin-gonic/gin"
+  "net/http"
```
ahsu1230: packages with "github.com" are external packages. The others like net and strconv are built-in go packages. For now, I would like to keep them separated so could you keep net and strconv together on top, then external packages on the bottom please?

orion/controllers/programs/programsController.go
```diff
@@ -81,6 +79,40 @@ func DeleteProgram(c *gin.Context) {
   return
 }
 
-func CheckValidProgram() bool {
+func CheckValidProgram(c *gin.Context) bool {
```
ahsu1230: nice try!
`gin` is an external package for HTTP request and response handling. We'll talk more about that later.

Instead, the input here should be a Program. We are going to use function CheckValidProgram() to check a passed in Program. If it's valid we return true, otherwise we return false.

orion/controllers/programs/programsController.go
```diff
@@ -33,13 +34,10 @@ func CreateProgram(c *gin.Context) {
...
 
-  // TODO: implement for real!
-  // isValid := CheckValidProgram(newProgram);
-  // if (isValid) {
-  //
-  // } else {
-  //
-  // }
+  // TODO: test if this works
+  if CheckValidProgram(c) == false {
```
ahsu1230: Look below about my comments regarding `gin`. You won't need it yet.
Instead, the parameter here should be a `Program`.

orion/controllers/programs/programsController.go
```diff
@@ -81,6 +79,40 @@ func DeleteProgram(c *gin.Context) {
...
+
+  // Checks if program name is empty
+  if name == "" {
+    c.String(http.StatusBadRequest, "Invalid name " + name)
+    return false
+  }
+
+  // Checks if the program name is a string only
+  for _, i := range name {
+    if (i < 'a' || i > 'z') && (i < 'A' || i > 'Z') && i != '_' {
```
ahsu1230: good check. But this program name could also have numbers as well. I wonder if we can use a regex (regular expression) to do this check? Have you ever done one of these before?

orion/controllers/programs/programsController.go
```diff
@@ -81,6 +79,40 @@ func DeleteProgram(c *gin.Context) {
...
+      c.String(http.StatusBadRequest, "Invalid name " + name)
```
ahsu1230: since we're not using gin, you won't need these c.String(...). I'll tell you later what these are for.

orion/controllers/programs/programsController.go
```diff
@@ -81,6 +79,40 @@ func DeleteProgram(c *gin.Context) {
...
+      return false
+    }
+  }
+
+  // Checks if the grades are integers
+  first, err1 := strconv.Atoi(grade1)
+  second, err2 := strconv.Atoi(grade2)
+
+  if err1 != nil || err2 != nil {
```
ahsu1230: nice!

orion/controllers/programs/programsController.go
```diff
@@ -81,6 +79,40 @@ func DeleteProgram(c *gin.Context) {
...
+  // Checks if the program name is a string only
+  for _, i := range name {
+    if (i < 'a' || i > 'z') && (i < 'A' || i > 'Z') && i != '_' {
+      c.String(http.StatusBadRequest, "Invalid name " + name)
+      return false
+    }
+  }
+
+  // Checks if the grades are integers
+  first, err1 := strconv.Atoi(grade1)
```
ahsu1230: maybe a better variable name than `first` or `second` could be `gradeNum1` and `gradeNum2`. Something to convey that the grades have turned into numbers.

orion/controllers/programs/programsController.go
```diff
@@ -81,6 +79,40 @@ func DeleteProgram(c *gin.Context) {
...
+  second, err2 := strconv.Atoi(grade2)
+
+  if err1 != nil || err2 != nil {
+    c.String(http.StatusBadRequest, "Invalid grades " + grade1 + " to " + grade2)
+    return false
+  }
+
+  // Checks if the grades are valid
+  if !(first <= second && first >= 1 && second <= 12) {
```
ahsu1230: nice!

orion/controllers/programs/programs_test.go
```diff
@@ -12,8 +12,9 @@ func TestExample(t *testing.T) {
 }
 
 func TestCheckProgram(t *testing.T) {
```
ahsu1230: I put this as a placeholder... but maybe we should separate this big mega test into 2 different tests. You just wrote code to check a Program's name and grades.
So let's have one unit test be all about checking program names `TestCheckProgramNames(...)`.
And another unit test be all about checking program grades `TestCheckProgramGrades(...)`.
In each unit test, you can then do a deep dive on checking all edges cases for the respective test.

ahsu1230:
Overall, very good try! I hope my insights helped.
As for testing things out, I'd say just think of good unit tests and make sure those pass. We will talk about `gin` stuff tomorrow!

@mzhang2021 If you're still working on your stuff, take a look here to see how you can improve your code. If you have any questions, leave some comments and maybe @tonywu315 or I can take a look!

orion/controllers/programs/programsController.go
tonywu315: I’ve done regex before. If it can include numbers, then is this check necessary? I thought the variable was already a string.

ahsu1230: I think it's still worth it. We want to check all alphanumeric character values and underscores. no special characters like -.?*^ etc.

orion/controllers/programs/programsController.go
```diff
@@ -81,6 +84,42 @@ func DeleteProgram(c *gin.Context) {
...
+  // Checks if program name or program ID is empty
+  if programId == "" {
+    return errors.New("empty Program ID")
+  }
+
+  // TODO: use regex
+  // Checks if the program ID is alphanumeric
+  for _, i := range name {
+    if (i < 'a' || i > 'z') && (i < 'A' || i > 'Z') && (i < '1' || i > '0') && i != '_' {
+      return errors.New("invalid program name")
```
ahsu1230: invalid program id

orion/controllers/programs/programsController.go
```diff
@@ -81,6 +84,42 @@ func DeleteProgram(c *gin.Context) {
...
+  }
+
+  if name == "" {
+    return errors.New("empty program name")
+  }
+
+  // TODO: Check if program name is alphanumeric and can include spaces, underscores, and ampersands
+  // Checks if the program name is alphanumeric
+  for _, i := range name {
+    if (i < 'a' || i > 'z') && (i < 'A' || i > 'Z') && (i < '1' || i > '0') && i != '_' {
```
ahsu1230: you can delete this for loop and let max handle it. One is to avoid merge conflicts later. And two is because program name will have less restrictions.

orion/controllers/programs/programsController.go
```diff
@@ -81,6 +84,42 @@ func DeleteProgram(c *gin.Context) {
...
+  name := program.Name
+  grade1 := program.Grade1
+  grade2 := program.Grade2
+
+  // Checks if program name or program ID is empty
+  if programId == "" {
+    return errors.New("empty Program ID")
+  }
+
+  // TODO: use regex
```
ahsu1230: remove TODO

orion/controllers/programs/programsController.go
```diff
@@ -81,6 +84,42 @@ func DeleteProgram(c *gin.Context) {
...
-func CheckValidProgram() bool {
-  return true
+func CheckValidProgram(program Program) error {
+  // Retrieves the inputted values
+  programId := program.ProgramId
+  name := program.Name
+  grade1 := program.Grade1
+  grade2 := program.Grade2
+
+  // Checks if program name or program ID is empty
```
ahsu1230: You can remove the comments here. This part is checking if ProgramId is empty which is self-explanatory from the code

ahsu1230:
Also CircleCI (the automatic checks in github) is working now. If you click on `Details` next to "Your tests failed on CircleCI", you'll be able to see what went wrong with your test/build.

orion/controllers/programs/programs_test.go
```diff
@@ -11,9 +11,78 @@ func TestExample(t *testing.T) {
...
+
+  program = Program{ProgramId: "", Name: "Test Program", Grade1: 1, Grade2: 12, Description: "Description"}
+  err = CheckValidProgram(program)
+  if err == nil {
+    t.Errorf("Check was incorrect, got: nil, expected: invalid program id")
   }
 }
+
+func TestValidGrades(t *testing.T) {
+  program := Program{ProgramId: "test_program", Name: "Test Program", Grade1: 1, Grade2: 12, Description: "Description"}
```
ahsu1230: It seems like we are creating a new Program object for every subtest.
How about at the beginning of the test we create a "base Program", and then for every subtest, we only change the `grade1` and `grade2` fields?

So you can have:
```
program := Program{...}
...

program.Grade1 = 2
program.Grade2 = 3
...
program.Grade1 = 12
program.Grade2 = 0
```
Otherwise, lots of code redundancy :)

orion/controllers/programs/programs_test.go
```diff
@@ -11,9 +11,78 @@ func TestExample(t *testing.T) {
   }
 }
 
-func TestCheckProgram(t *testing.T) {
-  isValid := CheckValidProgram()
-  if (isValid == false) {
-    t.Errorf("Check was incorrect, got: %t, expected: %t.", isValid, false)
+func TestValidProgramId(t *testing.T) {
+  program := Program{ProgramId: "test_program", Name: "Test Program", Grade1: 1, Grade2: 12, Description: "Description"}
```
ahsu1230: Look at my comment under the grade test.

orion/controllers/programs/programsController.go
```diff
@@ -81,6 +85,24 @@ func DeleteProgram(c *gin.Context) {
...
+  programId := program.ProgramId
+  grade1 := program.Grade1
+  grade2 := program.Grade2
+
+  if programId == "" {
+    return errors.New("empty Program ID")
+  }
+
+  // Checks if the program ID is in the form of alphanumeric strings separated by underscores
+  if matches, _ := regexp.MatchString("^[[:alnum:]]+(_[[:alnum:]]+)*$", programId); !matches {
```
ahsu1230: > "^[[:alnum:]]+(_[[:alnum:]]+)*$"

Can you make this a global variable?
Every time we call this function, we re-create this String.
Instead, with a global variable, it is created once and simply always referenced.

At the top of the file, you could do:
```
REGEX_PROGRAM_ID := "..."
```

I like all upper-cased and snake case here because it signals this variable should NOT be changed and it's a global variable. Variables with camelCase imply a "regular" variable.

orion/controllers/programs/programsController.go
```diff
@@ -1,10 +1,15 @@
 package programs
 
+
```
ahsu1230: extra line here

PR #15, opened 2020-02-08
Add MySql and curl

ahsu1230:
@tonywu315 @mzhang2021

### DIFF
--- a/resources/backend/README_3_sql.md
+++ b/resources/backend/README_3_sql.md
@@ -58,9 +58,9 @@ SELECT * FROM persons;
 The above line selects ALL rows in table `persons`.
 If we wanted to select particular rows or columns, we would use the `WHERE` clause like so:
 ```
-SELECT * FROM persons WHERE id=1;
+SELECT * FROM persons WHERE person_id=1;
 SELECT * FROM persons WHERE first_name="Ash";
-SELECT address, city FROM persons WHERE last_name="Ketchum"
+SELECT address, city FROM persons WHERE last_name="Ketchum";
 ```
 
 To update certain rows, you can use this:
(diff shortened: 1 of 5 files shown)


resources/backend/README_3_sql.md
```diff
@@ -1,6 +1,87 @@
...
+Add 4 more rows with varying values to get more data into our database!
+Once you are finished, you can use this command to see all of them:
+```
+SELECT * FROM persons;
+```
+
+The above line selects ALL rows in table `persons`.
+If we wanted to select particular rows or columns, we would use the `WHERE` clause like so:
+```
+SELECT * FROM persons WHERE id=1;
```
tonywu315: SELECT * FROM persons WHERE person_id=1;

resources/backend/README_3_sql.md
```diff
@@ -1,6 +1,87 @@
...
+SELECT * FROM persons WHERE first_name="Ash";
+SELECT address, city FROM persons WHERE last_name="Ketchum"
```
tonywu315: semicolon at the end

resources/backend/README_3_sql.md
```diff
@@ -1,6 +1,87 @@
...
+```
+After playing around with these commands, use `SELECT * FROM persons;` to see how your table looks after your changes.
+
+When you are finished, you can use the following commands to:
+ - delete table `persons`
+ - delete database `testDb`
+ - quit out of MySQL Terminal/Shell
+```
+DROP TABLE persons;
+DROP testDb;
```
ahsu1230: DROP SCHEMA testDb;
DROP DATABASE testDb;

PR #18, opened 2020-02-08
Added validation for description and name, shortened existing validation code

mzhang2021:
@ahsu1230 
@tonywu315

### DIFF
--- a/orion/controllers/programs/programsController.go
+++ b/orion/controllers/programs/programsController.go
@@ -33,13 +33,10 @@ func CreateProgram(c *gin.Context) {
   var programJson Program
   c.BindJSON(&programJson)
 
-  // TODO: implement for real!
-  // isValid := CheckValidProgram(newProgram);
-  // if (isValid) {
-  //
-  // } else {
-  //
-  // }
+  if CheckValidProgram(programJson) == false {
+    //c.String(http.StatusBadRequest)
+    return
+  }
 
   // Query Repo (INSERT & SELECT)
   err := InsertProgram(programJson)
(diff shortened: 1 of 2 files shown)


orion/controllers/programs/programsController.go
```diff
@@ -81,6 +84,29 @@ func DeleteProgram(c *gin.Context) {
   return
 }
 
-func CheckValidProgram() bool {
+func CheckValidProgram(program Program) bool {
```
ahsu1230: So at the end of our session, Tony and I talked about how instead of returning a `bool`, `CheckValidProgram(...)` will now return an `error` type. Booleans funnel results into two values, but it would be good to know why a program is not valid. Take a look at his change to see how to use `error`.

orion/controllers/programs/programsController.go
```diff
@@ -81,6 +84,29 @@ func DeleteProgram(c *gin.Context) {
...
+  grade2 := program.Grade2
+  description := program.Description
+
+  // Checks if variables aren't empty
+  if programId == 0 || name == "" || description == "" {
+    return false
+  }
+
+  // Name validation
+  match, _ := regexp.MatchString(`^[A-Z]\S*(\s[A-Z]\S*)*$`, s)
```
ahsu1230: Nice!

orion/controllers/programs/programs_test.go
```diff
@@ -12,8 +12,9 @@ func TestExample(t *testing.T) {
 }
 
 func TestCheckProgram(t *testing.T) {
-  isValid := CheckValidProgram()
-  if (isValid == false) {
-    t.Errorf("Check was incorrect, got: %t, expected: %t.", isValid, false)
+  program := Program{Name: "Test Program", Grade1: 1, Grade2: 12}
+  isValid := CheckValidProgram(program)
+  if !isValid == false {
+    t.Errorf("Check was incorrect, got: %t, expected: %t.", isValid, true)
```
ahsu1230: Can we get more unit tests?
You've tested two words. What if it was three words?
What if there was one word?
Can you also test failure cases as well?

PR #27, opened 2020-02-14
Fixed spacing

tonywu315:


orion/database/db.go
```diff
@@ -1,66 +1,66 @@
 package database
 
 import (
-  "database/sql"
-  "fmt"
+	"database/sql"
+	"fmt"
 
-  "github.com/jmoiron/sqlx"
-  _ "github.com/go-sql-driver/mysql"
```
ahsu1230: we need to keep these `_`.
These underscores are like "partial" imports. They tell the program to import what it needs as opposed to the whole library.

orion/database/db.go
```diff
@@ -1,66 +1,66 @@
...
-  "github.com/golang-migrate/migrate"
-  "github.com/golang-migrate/migrate/database/mysql"
-  _ "github.com/golang-migrate/migrate/source/file"
```
ahsu1230: same here
The indentation should look like:
```
"github.com/...
 _ "github.com/...."
```
So the second line should be: (space) _ (space) "github.com/..."

orion/database/db.go
ahsu1230: Does IntelliJ remove these?

PR #29, opened 2020-02-16
Announcements

mzhang2021:
@ahsu1230 
@tonywu315 
Just want to check in and see if I'm on the right track.
TODO: The API endpoint stuff
UPDATE: Finished


orion/controllers/.project
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ahsu1230: I know you added the `.gitignore` file last time. I think you should change it to this: `**/*/.project`

orion/controllers/announce/announceModel.go
```diff
@@ -0,0 +1,15 @@
...
+
+type Announce struct {
+	Id				uint
+	CreatedAt		uint
+	UpdatedAt		uint
+	DeletedAt		sql.NullInt64
+	AnnounceId		string
+	Title			string
+	Message			string
+}
```
ahsu1230: This might be my mistake because I did not clearly communicate this:
The Announce struct should follow what the database columns look like.
In mysql, if you ` use mathnavdb` and then `describe announcements`, you'll see its properties:
```
+------------+------------------+------+-----+---------+----------------+
| Field      | Type             | Null | Key | Default | Extra          |
+------------+------------------+------+-----+---------+----------------+
| id         | int(10) unsigned | NO   | PRI | NULL    | auto_increment |
| created_at | bigint(20)       | NO   |     | NULL    |                |
| updated_at | bigint(20)       | NO   |     | NULL    |                |
| deleted_at | datetime         | YES  |     | NULL    |                |
| posted_at  | bigint(20)       | NO   |     | NULL    |                |
| author     | varchar(255)     | NO   |     | NULL    |                |
| message    | text             | NO   |     | NULL    |                |
+------------+------------------+------+-----+---------+----------------+
```

Here, the `Id` is just the row number and we do not need an `announceId`. Unlike `programs` where all the endpoints API use the String `programId`, for announcements, we will just use `id` (unsigned integer).
In addition, we won't need a `Title` (yet) and we'll need a `postedAt` and `author` column.
Sorry for the miscommunication.

orion/controllers/announce/announceController.go
```diff
@@ -0,0 +1,115 @@
...
+
+import (
+	"errors"
+	"net/http"
+	"regexp"
+	
+	"github.com/gin-gonic/gin"
+)
+
+const REGEX_ANNOUNCE_ID = `^[[:alnum:]]+(_[[:alnum:]]+)*$`
```
ahsu1230: Thank you for leaving these checks. `AnnounceId` won't be used. Instead we will just be using the Row `Id`

orion/controllers/announce/announceController.go
```diff
@@ -0,0 +1,115 @@
...
+const REGEX_TITLE = `^[A-Z0-9][[:alnum:]-]*([- _]([(]?#\d[)]?|&|([(]?[[:alnum:]]+[)]?)))*$`
```
ahsu1230: I appreciate this. For now, we can keep things simple and just allow non-empty strings for message, author, etc.

orion/controllers/announce/announceRepo.go
```diff
@@ -0,0 +1,60 @@
...
+func InsertAnnouncement(announce Announce) (error) {
+	now := utils.TimestampNow()
+	sqlStatement := "INSERT INTO announcements " +
+		"(created_at, updated_at, deleted_at, announce_id, title, message) " +
+        "VALUES (:createdAt, :updatedAt, :deletedAt, :announceId, :title, :message)"
+	parameters := map[string]interface{} {
+		"createdAt": now,
+		"updatedAt": now,
+		"deletedAt": nil,
+		"announceId": announce.AnnounceId,
```
ahsu1230: There's going to be a `postedAt` column. This one is different because you can create an announcement, but not post it yet (like schedule an announcement in the future). This parameter will be an epoch time in milliseconds (so you may assume it will be unsigned int in the `Announce` object).

orion/controllers/announce/announceRepo.go
```diff
@@ -0,0 +1,60 @@
...
+	return err
+}
+
+func UpdateAnnouncementById(oldAnnounceId string, announce Announce) (error) {
+	now := utils.TimestampNow()
+
+	sqlStatement := "UPDATE announcements SET " +
+		"updated_at=:updatedAt, " +
+		"announce_id=:announceId, " +
+		"title=:grade1, " +
```
ahsu1230: you have `grade1` and `grade2` here.

orion/controllers/announce/announce_test.go
```diff
@@ -0,0 +1,122 @@
...
+		t.Errorf("Check was incorrect, got: %s, expected: nil", err.Error())
+	}
+
+	// Failure examples
+	announce.Message = ""
+	if err := CheckValidAnnouncement(announce); err == nil {
+		t.Error("Check was incorrect, got: nil, expected: empty message")
+	}
+	
+	announce.Message = "                                                    "
```
ahsu1230: Nice, I think it's good to guard against this. So maybe disallow empty strings and strings with no alphabetic letters? This goes for both messages and author.

orion/controllers/.project
mzhang2021: Yeah I think my previous commits somehow got caught in this branch. The .gitignore change was not intentional.

orion/controllers/announce/announceController.go
```diff
@@ -0,0 +1,107 @@
...
+	c.BindJSON(&announceJson)
+
+	if err := CheckValidAnnouncement(announceJson); err != nil {
+		c.String(http.StatusBadRequest, err.Error())
+		return
+	}
+
+	// Query Repo (INSERT & SELECT)
+	err := InsertAnnouncement(announceJson)
+	if err != nil {
```
ahsu1230: if it looks better, you can do this:
```
if err := InsertAnnouncement(announceJson); err != nil {
}
```
It shortens the code by a line and we don't have a lingering `err` variable.

orion/controllers/announce/announceModel.go
```diff
@@ -0,0 +1,15 @@
+package announce
+
+import (
+	"database/sql"
+)
+
+type Announce struct {
+	Id				string			`db:"id"`
```
ahsu1230: indentation issue?

orion/controllers/announce/announceRepo.go
```diff
@@ -0,0 +1,60 @@
...
+		"deletedAt": nil,
+		"postedAt": announce.PostedAt,
+		"author": announce.Author,
+		"message": announce.Message,
+	}
+	_, err := database.DbSqlx.NamedExec(sqlStatement, parameters)
+	return err
+}
+
+func UpdateAnnouncementById(id string, announce Announce) (error) {
```
ahsu1230: id here is a uint

orion/controllers/announce/announceRepo.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/ahsu1230/mathnavigatorSite/orion/database"
+)
+
+func GetAllAnnouncements() []Announce {
+	var announceList []Announce
+	database.DbSqlx.Select(&announceList, "SELECT * FROM announcements")
+	return announceList
+}
+
+func GetAnnouncementById(id string) (Announce, error) {
```
ahsu1230: I think id is supposed to be a uint

orion/controllers/announce/announceController.go
```diff
@@ -0,0 +1,107 @@
...
+		panic(err)
+	} else {
+		c.JSON(http.StatusOK, nil)
+	}
+	return
+}
+
+func UpdateAnnouncement(c *gin.Context) {
+	// Incoming JSON & Parameters
+	id := c.Param("id")
```
ahsu1230: `id` here is a parameter from the URL so it will first be a string.
I think we will have to parse this string and convert it into a uint.

orion/controllers/announce/announceController.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"errors"
+	"net/http"
+	"regexp"
+	
+	"github.com/gin-gonic/gin"
+)
+
+const REGEX = `[A-Za-z]+`
```
ahsu1230: can you give a more meaningful REGEX here and a comment of what it is trying to regex?
REGEX_MESSAGE and REGEX_AUTHOR?
REGEX_LETTERS_ONLY?
REGEX_NON_EMPTY?

orion/controllers/programs/programsController.go
```diff
@@ -9,6 +9,7 @@ import (
 
 const REGEX_PROGRAM_ID = `^[[:alnum:]]+(_[[:alnum:]]+)*$`
```
ahsu1230: I shouldn't mentioned before. Could we get a comment per program id, name and description?
Basically it would be nice to have a summary of what these REGEX are checking.

orion/main.go
```diff
@@ -48,7 +49,14 @@ func main() {
...
   // apiLocations := router.Group("api/locations/")
-  // apiAnnounce := router.Group("api/announce/")
+  apiAnnounce := router.Group("api/announce/")
+  {
+  	apiAnnounce.GET("/v1/all", announce.GetAnnouncements)
+  	apiAnnounce.POST("/v1/create", announce.CreateAnnouncement)
+  	apiAnnounce.GET("/v1/announce/:id", announce.GetAnnouncement)
+  	apiAnnounce.POST("/v1/announce/:id", announce.UpdateAnnouncement)
+  	apiAnnounce.DELETE("/v1/announce/:id", announce.DeleteAnnouncement)
+  }
```
ahsu1230: very nice! :)

orion/controllers/announce/announceRepo.go
ahsu1230: Going forward, we should expect `id` to always mean row ids and should always be unsigned integers.
anything else like program_id, location_id, etc. might be something else like a string.

orion/controllers/announce/announceController.go
mzhang2021: Yeah I decided to make id a string so I wouldn't have to parse and convert the string, but if it's better practice to make id type uint then I'll do that instead.

orion/controllers/announce/announceModel.go
mzhang2021: I think that's just github displaying default 8 space tabbing, it looks more normal with 4 space tabbing offline.
Capture

orion/controllers/.project
ahsu1230: bump

orion/main.go
```diff
@@ -8,62 +8,70 @@ import (
...
+  {
+    apiPrograms.GET("/v1/all", programs.GetPrograms)
+    apiPrograms.POST("/v1/create", programs.CreateProgram)
+    apiPrograms.GET("/v1/program/:programId", programs.GetProgram)
+    apiPrograms.POST("/v1/program/:programId", programs.UpdateProgram)
+    apiPrograms.DELETE("/v1/program/:programId", programs.DeleteProgram)
+  }
+  // apiClasses := router.Group("api/classes/")
+  // apiLocations := router.Group("api/locations/")
+  apiAnnounce := router.Group("api/announce/")
```
ahsu1230: Miscommunication on my part. Change this back to `api/announcements` here.
For the other endpoints below, since we are operating on one announcement at a time, those should stay as `/v1/announce/...`.

orion/controllers/announce/announceController.go
```diff
@@ -0,0 +1,107 @@
...
+	} else {
+		c.Status(http.StatusOK)
+	}
+	return
+}
+
+func DeleteAnnouncement(c *gin.Context) {
+	// Incoming Parameters
+	id64, _ := strconv.ParseUint(c.Param("id"), 10, 32)
+	id := uint(id64)
```
ahsu1230:
```
id64, _ := strconv.ParseUint(c.Param("id"), 10, 32)
id := uint(id64)
```

You do this part here a few times in this file. Maybe we can refactor this part out to a function in this file so you don't have to keep repeating it.

PR #30, opened 2020-02-20
Achievements

tonywu315:


orion/controllers/achieve/achieveController.go
```diff
@@ -0,0 +1,108 @@
...
+	achieveList := GetAllAchievements()
+
+	// JSON Response
+	c.JSON(http.StatusOK, achieveList)
+	return
+}
+
+func GetAchievement(c *gin.Context) {
+	// Incoming parameters
+	id64, _ := strconv.ParseUint(c.Param("id"), 10, 32)
```
ahsu1230: nice! Though I think if we get an error, we should probably do something about it instead of continuing.
For now, you can use `panic(err)`.

We will probably do something better to handle errors in the near future.

orion/controllers/achieve/achieveController.go
```diff
@@ -0,0 +1,108 @@
...
+	return
+}
+
+func CheckValidAchievement(achieve Achieve) error {
+	// Retrieves the inputted values
+	year := achieve.Year
+	message := achieve.Message
+
+	// Year validation
+	if year < 2000 {
```
ahsu1230: nice I think this is good enough for now. At least it catches if year = 0.

orion/main.go
```diff
@@ -1,62 +1,71 @@
 package main
+
 import (
-  "fmt"
+	"fmt"
```
ahsu1230: thanks for the re-formatting!

orion/controllers/achieve/achieveController.go
```diff
@@ -0,0 +1,108 @@
...
+		return errors.New("invalid year")
```
ahsu1230: Let's make error messages all start with a Capital letter for just the first word.
`Invalid year`

orion/controllers/achieve/achieveController.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	// Message validation
+	if message == "" {
+		return errors.New("invalid message")
```
ahsu1230: same here

orion/controllers/achieve/achieveController.go
tonywu315: The go specifications say that error strings should not be capitalized.

tonywu315: The Go specifications say that error strings should be lowercase @ahsu1230

ahsu1230: oh You're right. nvm
https://github.com/golang/go/wiki/Errors

orion/controllers/achieve/achieveController.go
```diff
@@ -0,0 +1,108 @@
...
+		panic(err)
+	} else {
+		c.JSON(http.StatusOK, nil)
+	}
+	return
+}
+
+func UpdateAchievement(c *gin.Context) {
+	// Incoming JSON & Parameters
+	id64, _ := strconv.ParseUint(c.Param("id"), 10, 32)
```
ahsu1230: I think this should be `achieveId`

orion/controllers/achieve/achieveController.go
```diff
@@ -0,0 +1,108 @@
...
+		panic(err)
+	} else {
+		c.JSON(http.StatusOK, nil)
+	}
+	return
+}
+
+func DeleteAchievement(c *gin.Context) {
+	// Incoming Parameters
+	id64, _ := strconv.ParseUint(c.Param("id"), 10, 32)
```
ahsu1230: same here

orion/main.go
```diff
@@ -1,62 +1,71 @@
...
+		apiPrograms.GET("/v1/all", programs.GetPrograms)
+		apiPrograms.POST("/v1/create", programs.CreateProgram)
+		apiPrograms.GET("/v1/program/:programId", programs.GetProgram)
+		apiPrograms.POST("/v1/program/:programId", programs.UpdateProgram)
+		apiPrograms.DELETE("/v1/program/:programId", programs.DeleteProgram)
+	}
+	// apiClasses := router.Group("api/classes/")
+	// apiLocations := router.Group("api/locations/")
+	// apiAnnounce := router.Group("api/announce/")
+	apiAchieve := router.Group("api/achieve/")
```
ahsu1230: can we make this "api/achievements" for its full name?
The rest can stay as "v1/achieve/..."

orion/controllers/achieve/achieveController.go
```diff
@@ -0,0 +1,117 @@
...
+		panic(err)
+	}
+	id := uint(id64)
+
+	// Query Repo (DELETE)
+	err = DeleteAchievementById(id)
+	if err != nil {
+		panic(err)
+	} else {
+		c.String(http.StatusOK, "Deleted Achieve " + strconv.FormatUint(id64, 10))
```
ahsu1230: Let's change this to `c.Status(http.StatusOK)`

orion/main.go
ahsu1230: can we turn this into `api/achievements/`?

orion/controllers/achieve/achieveController.go
```diff
@@ -0,0 +1,111 @@
...
+	// Query Repo
+	achieveList := GetAllAchievements()
+
+	// JSON Response
+	c.JSON(http.StatusOK, achieveList)
+	return
+}
+
+func GetAchievement(c *gin.Context) {
+	id := StringToUint(c)
```
ahsu1230: I think a more descriptive function name is better here.
```
id := parseParamId(c)
```

Since this is a private function, you can make it lowercase.
Upper case functions indicate that this function will be used by other packages.

PR #33, opened 2020-02-23
Semesters

tonywu315:
Created semesters backend flow
@ahsu1230 @mzhang2021


orion/main.go
```diff
@@ -59,7 +60,14 @@ func main() {
 		apiAchieve.POST("/v1/achievements/:id", achieve.UpdateAchievement)
 		apiAchieve.DELETE("/v1/achievements/:id", achieve.DeleteAchievement)
 	}
-	// apiSemesters := router.Group("api/semesters/")
+	apiSemesters := router.Group("api/semesters/")
+	{
+		apiSemesters.GET("/v1/all", semesters.GetSemesters)
+		apiSemesters.POST("/v1/create", semesters.CreateSemester)
+		apiSemesters.GET("/v1/semesters/:id", semesters.GetSemester)
```
ahsu1230: The group should stay as `/api/semesters/`
The GET, POST and DELETE here should indicate we are operating on one `semester`.
So can you change the following to `v1/semester/:id`

orion/main.go
```diff
@@ -59,7 +60,14 @@ func main() {
 		apiAchieve.POST("/v1/achievements/:id", achieve.UpdateAchievement)
```
ahsu1230: I missed it last time, could you do the same here?

orion/database/migrations/000004_create_table_semesters.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE semesters
+(
+    id          int unsigned NOT NULL AUTO_INCREMENT,
+    created_at  bigint(20)   NOT NULL,
+    updated_at  bigint(20)   NOT NULL,
+    deleted_at  datetime,
+    semester_id text         NOT NULL UNIQUE,
```
ahsu1230: semester_ids are short, so instead of text, make it a varchar(64) as well, I think.

orion/controllers/semesters/semestersController.go
```diff
@@ -0,0 +1,109 @@
...
+
+import (
+	"errors"
+	"net/http"
+	"regexp"
+
+	"github.com/gin-gonic/gin"
+)
+
+const REGEX_SEMESTER_ID = `^[1-9]\d{3,}_((spring)|(summer)|(fall)|(winter))$`
```
ahsu1230: Can you add comments above these REGEX expressions to briefly explain what the REGEX is checking?

orion/controllers/semesters/semestersController.go
```diff
@@ -0,0 +1,109 @@
...
+		c.String(http.StatusBadRequest, err.Error())
+		return
+	}
+
+	// Query Repo (INSERT & SELECT)
+	err := InsertSemester(semesterJson)
+	if err != nil {
+		panic(err)
+	} else {
+		c.JSON(http.StatusOK, nil)
```
ahsu1230: if no JSON, let's just use a c.Status(OK)

orion/controllers/semesters/semestersController.go
```diff
@@ -0,0 +1,109 @@
...
+		c.String(http.StatusBadRequest, err.Error())
+		return
+	}
+
+	// Query Repo (UPDATE & SELECT)
+	err := UpdateSemesterById(semesterId, semesterJson)
+	if err != nil {
+		panic(err)
+	} else {
+		c.JSON(http.StatusOK, nil)
```
ahsu1230: same here, just use a c.Status(_)

orion/controllers/semesters/semesters_test.go
```diff
@@ -0,0 +1,49 @@
...
+	semester := Semester{SemesterId: "2020_spring", Title: "Test title"}
+	if err := CheckValidSemester(semester); err != nil {
+		t.Errorf("Check was incorrect, got: %s, expected: nil", err.Error())
+	}
+
+	semester.Title = "Winter 2222 (#1 class)"
+	if err := CheckValidSemester(semester); err != nil {
+		t.Errorf("Check was incorrect, got: %s, expected: nil", err.Error())
+	}
+
```
ahsu1230: Comment checks for incorrect semester names

orion/controllers/semesters/semesters_test.go
```diff
@@ -0,0 +1,49 @@
...
+	semester.SemesterId = "2050_winter"
+	if err := CheckValidSemester(semester); err != nil {
+		t.Errorf("Check was incorrect, got: %s, expected: nil", err.Error())
+	}
+
+	semester.SemesterId = "0999_spring"
```
ahsu1230: Comment here about checking invalid semester ids

.gitignore
```diff
@@ -8,5 +8,6 @@
 **/*/go.sum
 **/.project
```
ahsu1230: thanks for deleting .project. Can you also remove this line here `**/.project` but leave the other one

orion/controllers/announce/announceController.go
```diff
@@ -33,7 +34,7 @@ func GetAnnouncement(c *gin.Context) {
 	if _, err := GetAnnouncementById(id); err != nil {
 		panic(err)
 	} else {
-		c.Status(http.StatusOK)
+		c.Status(http.StatusNoContent)
```
ahsu1230: uh oh, I think this is a mistake. This function is supposed to return one singular Announcement object.
So, this should be `c.Json(http.StatusOK, announceObj)`

ahsu1230: I should've caught this. Argh I can't wait until I finish setting up a better unit test framework.

orion/controllers/semesters/semestersController.go
```diff
@@ -0,0 +1,112 @@
...
+
+import (
+	"errors"
+	"net/http"
+	"regexp"
+
+	"github.com/gin-gonic/gin"
+)
+
+// In the form year_season
```
ahsu1230: In the form "year_season" i.e. 2020_fall

orion/main.go
```diff
@@ -55,19 +56,26 @@ func main() {
 	{
 		apiAnnounce.GET("/v1/all", announce.GetAnnouncements)
 		apiAnnounce.POST("/v1/create", announce.CreateAnnouncement)
-		apiAnnounce.GET("/v1/announce/:id", announce.GetAnnouncement)
-		apiAnnounce.POST("/v1/announce/:id", announce.UpdateAnnouncement)
-		apiAnnounce.DELETE("/v1/announce/:id", announce.DeleteAnnouncement)
+		apiAnnounce.GET("/v1/announcement/:id", announce.GetAnnouncement)
+		apiAnnounce.POST("/v1/announcement/:id", announce.UpdateAnnouncement)
+		apiAnnounce.DELETE("/v1/announcement/:id", announce.DeleteAnnouncement)
```
ahsu1230: nice call, I think this is better.

orion/controllers/semesters/semestersModel.go
```diff
@@ -0,0 +1,14 @@
+package semesters
```
ahsu1230: nit: `semesterModel.go`

orion/controllers/announce/announceController.go
mzhang2021: Hmm I think think it used to say c.Json, but then you told me to change it to c.Status.

ahsu1230: ah I see. So if we are in the situation where we return StatusOK and no other content, we can simply just use Status(NoContent).
On the other hand, if we are returning additional values (in this case a program object), we should use c.Status(OK, object).

orion/controllers/semesters/semestersModel.go
tonywu315: Why is the package also semestersModel.go

PR #37, opened 2020-02-26
updated announce.styl

MiniViolin:


orion/sites/admin/src/announce/announce.styl
```diff
@@ -1,13 +1,48 @@
 @import "../static.styl"
 
+
```
ahsu1230: extra line

orion/sites/admin/src/announce/announce.styl
```diff
@@ -1,13 +1,48 @@
...
+  font-family: fontFamilyRoboto /*not working*/
 
   & h1
     margin: spacing4x 0
     line-height: lineHeight3
     font-size: fontSize3
+  & ul
+    padding: 0
+
+  .announce-lists
```
ahsu1230: If there is an indent here, please add a & before it.
```
#view-announce
    & .announce-lists
```

orion/sites/admin/src/announce/announce.styl
```diff
@@ -1,13 +1,48 @@
...
+
+    & .li-med
+      width: spacing16x
+      display: inline-block
+
+    & .li-large
+      width: spacing64x
+      display: inline-block
+
+  .announcement-button
```
ahsu1230: same here (add &)

orion/sites/admin/src/announce/announce.styl
```diff
@@ -1,13 +1,48 @@
...
+    display: inline-block
+    list-style: none
+    padding-left: 0
+    height: lineHeight6
+    line-height: lineHeight6
+    border-bottom: 1px solid colorGrayLight
+
+    & .li-small
+      width: spacing8x
+      display: inline-block
```
ahsu1230: there's a `display: inline-block` in here, .li-med, and .li-large.
Can we use a common className to give all these `.li-x` elements `display: inline-block` so it only appears once?

orion/sites/admin/src/announce/announce.styl
MiniViolin: Oh whoops, I'll do that for the next PR, but for the one I just pushed I hadn't seen this comment yet.

orion/sites/admin/src/announce/announce.styl
MiniViolin: I'm not sure I understand what you mean. Do you mean to indent the ul or and & and indent the .announce-lists?

orion/sites/admin/src/announce/announce.styl
MiniViolin: Originally I just had the display:inline-block in the .announce-lists class, which is the "parent" class for .li-med, .li-large, and .li-small. However, this causes formatting issues in that the display: inline-block doesn't appear in the list elements.

PR #41, opened 2020-02-29
Update static.styl with "lineHeight6"

MiniViolin:
And potentially other things but I'm pretty sure it was just lineHeight6


ahsu1230:
I think we can close this PR because this change is already in the other PR: https://github.com/ahsu1230/mathnavigatorSite/pull/42

PR #42, opened 2020-02-29
announce.js edits

MiniViolin:


orion/sites/admin/src/announce/announce.js
```diff
@@ -5,9 +5,62 @@ import ReactDOM from 'react-dom';
...
 		return (
       <div id="view-announce">
-        <h1>All Announcements</h1>
+        <h1>All Announcements ({numAnnouncements}) </h1>
+				<ul> {/* this is the subheading */}
+					<div className="announce-lists">
+						<li className="li-med"> Date </li>
+						<li className="li-med"> Author </li>
+						<li className="li-large"> Message </li>
+						<li className="li-small"> </li>
```
ahsu1230: is this line needed?

orion/sites/admin/src/announce/announce.js
```diff
@@ -5,9 +5,62 @@ import ReactDOM from 'react-dom';
...
+					</div>
+				</ul>
+
+				{/* these are the lists under the subheading */}
+				<ul>
+					<li className="announce-lists">
```
ahsu1230: I think maybe a better name could be "announce-list-item" or "announce-list-row"?
"announce-lists" implies multiple items or an entire list, when really you're only stying a single row.

orion/sites/admin/src/announce/announce.styl
```diff
@@ -5,9 +5,48 @@
...
+  & .announce-lists
+    display: inline-block
+    list-style: none
+    padding-left: 0
+    height: lineHeight6
+    line-height: lineHeight6
+    border-bottom: 1px solid colorGrayLight
+
+    & li
+      display: inline-block
```
ahsu1230: you might not need this. There are no `li` elements inside `.announce-lists`.

ahsu1230: Oh wait, looks like the subheader has `li` elements while the main list has `div` elements. Can you consolidate to one or the other? Can you make both the subheader and main list have either `li` elements or `div` elements? Up to you which way you want to go.

orion/sites/admin/src/announce/announce.js
```diff
@@ -5,9 +5,62 @@ import ReactDOM from 'react-dom';
...
+			date: "02/23/2020",
+			author: "Talkative",
+			message: "This will be a longer message later"
+		};
 		return (
       <div id="view-announce">
-        <h1>All Announcements</h1>
+        <h1>All Announcements ({numAnnouncements}) </h1>
+				<ul> {/* this is the subheading */}
+					<div className="announce-lists">
```
ahsu1230: Here, announce-lists is a `div` while the one below are `li`. Try to make them consistent. It will make your life easier in the future. Either way is fine, but as long as it is consistent. Here, maybe I'd lean to go towards with making "announce-lists" to `li`. And that's because it's the child of a `ul` parent.

Then, the children of "announce-lists" could either by a `ul` followed by many `li`s or simply a couple of `div`s.

orion/sites/admin/src/announce/announce.styl
```diff
@@ -5,9 +5,48 @@
...
+  & ul
+    padding: 0
+
+  & .announce-lists
+    display: inline-block
+    list-style: none
+    padding-left: 0
+    height: lineHeight6
+    line-height: lineHeight6
+    border-bottom: 1px solid colorGrayLight
```
ahsu1230: Do you happen to remember CSS property ordering for good style?

I think the order should be
 - display:
 - box-model stuff (width, height, margin, padding, border stuff)
 - text styles (fonts, color, background-color)
 - anything else
 - animation-related lines.

orion/sites/admin/src/announce/announce.styl
```diff
@@ -5,9 +5,48 @@
...
+
+  .announcement-button
+    display: inline-block
+    color: colorWhite
+    background-color: colorTurquoise
+    width: spacing32x
+    height: lineHeight6
+    line-height: lineHeight6
+    text-align: center
+    border-radius: spacing1x
```
ahsu1230: Look at the comment above about CSS property ordering
This can be re-ordered as:
```
display: inline-block
width: 
height
line-height
border-radius
text-align
color
background-color
```

The point of doing this is to easily see what properties are being styled on this class/element. When it's a random order, you'll probably miss a few lines and end up adding 2 of the same properties.
Another thing is the properties try to order themselves from most significant properties (display / position, etc.) to least significant properties (like colors). When the properties in a categorized order, they become much easier to debug!

orion/sites/admin/src/announce/announce.js
MiniViolin: "announce-lists" applies to each announcement, so each list item under the subheading. Should I still change it?

PR #47, opened 2020-03-07
Refactor

tonywu315:
Refactor for semester, achieve, and user.
@ahsu1230 
@mzhang2021 

I've only refactored the domains so far.


orion/main.go
```diff
@@ -1,43 +1,43 @@
 package main
 
 import (
-    "fmt"
-    "os"
-    "github.com/gin-contrib/cors"
-    "github.com/gin-gonic/gin"
+	"fmt"
```
ahsu1230: I think your IDE has tabs of 6 spaces. The standard is usually 4 spaces.

tonywu315: It's 4 spaces on my IDE

ahsu1230:
yikes this is a lot. I do appreciate you finishing all your tasks and trying to reformat code. But next time, if you have multiple tasks, put them into different branches and then create one pull request per branch.
For example,
- One PR for achievements refactor (branch name: `refactor_achievements`)
- One PR for semesters refactor (branch name: `refactor_achievements`)
- One PR for adding the user table (branch name: `feature_users`)
- One PR for refactoring regex (branch name: `refactor_regex`)
It might be annoying for you to have to switch between branches, but it's also even more annoying to have to fix this PR when problems arise. A large PR makes it harder for people to review, harder for you to pass tests, and harder for you to resolve merge conflicts if they arise.

orion/pkg/repos/migrations/000005_create_table_users.up.sql
```diff
@@ -0,0 +1,16 @@
...
+    id          int unsigned NOT NULL AUTO_INCREMENT,
+    created_at  datetime     NOT NULL,
+    updated_at  datetime     NOT NULL,
+    deleted_at  datetime,
+    first_name  varchar(32)  NOT NULL,
+    last_name   varchar(32)  NOT NULL,
+    middle_name varchar(32),
+    email       varchar(64)  NOT NULL,
+    phone       varchar(24)  NOT NULL,
+    isGuardian  boolean      NOT NULL,
```
ahsu1230: We can give this a `NOT NULL DEFAULT false` constraint.

orion/pkg/repos/migrations/000004_create_table_semesters.up.sql
```diff
@@ -1,11 +1,11 @@
...
 (
     id          int unsigned NOT NULL AUTO_INCREMENT,
-    created_at  bigint(20)   NOT NULL,
-    updated_at  bigint(20)   NOT NULL,
+    created_at  datetime     NOT NULL,
+    updated_at  datetime     NOT NULL,
     deleted_at  datetime,
     semester_id varchar(64)  NOT NULL UNIQUE,
-    title       varchar(64)  NOT NULL,
+    title       varchar(255) NOT NULL,
```
ahsu1230: I think a title will usually follow the format: `2019_fall` or `2025_summer`, so it shouldn't be too large. 255 characters will be overkill. The reason we want to keep this tight is because imagine for every row, we allocate 255 bytes of storage for the `title` field but only use like 12 bytes. That's a lot of wasted space and quickly racks up in databases. 

On that note, when deciding these amounts, we don't have to get crazy trying to optimize. But at the same time, we also want to give ourselves a good amount of leeway because changing this later can be painful.

orion/pkg/repos/migrations/000005_create_table_users.up.sql
ahsu1230: Sorry in MySQL, it is actually `NOT NULL DEFAULT 0`. 0 means false while 1 means true.

PR #49, opened 2020-03-07
reconfigure announce

mzhang2021:
@ahsu1230 
@tonywu315 
First commit is misnamed, was domains not controller


ahsu1230:
Hi @mzhang2021 , 
If you look closely at the error message, there's a nil pointer and it happens somewhere related to `AnnounceRepo`.

All you're missing is using the `Initialize()` function for `AnnounceRepo`. Since it's a repo interface, there are two places that it will be used (one for real functionality and one for testing).
For the real functionality, look at `repos_util.go` for the `SetupRepos()` function.
For the test initializing, look at `setup_test.go` under the `setupTestDatabase(...)` function. You should see a `// Initialize repos here` comment.

Once you have that, your tests should all pass. Great job!

ahsu1230:
Oh and another comment. The reason why your unit tests worked and not the integration test is because in the `_repo_tests`, we mock the database (with Datadog), so the db never actually gets called. Otherwise, it wouldn't be a unit test because we are testing our "repo" AND MySQL. So because we're not actually invoking database calls, it doesn't scream when you didn't initialize the db. In the `_repo_tests`, those unit tests you wrote only check if the correct SQL statements are getting invoked.

On the other hand, the integration tests actually spin up a MySQL environment and make real HTTP calls to itself so it'll actually test database connections and calls.
Hope that makes sense!

orion/pkg/controllers/announce_controller.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		panic(err)
+	}
+	return uint(id)
+}
+
+func GetAllAnnouncements(c *gin.Context) {
+	announceList, err := services.AnnounceService.GetAll()
+	if err != nil {
+		c.String(http.StatusInternalServerError, err.Error())
```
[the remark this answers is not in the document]
mzhang2021: err.Error() doesn't take any arguments.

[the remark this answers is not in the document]
ahsu1230: oh snap, sorry nevermind. The error will populate from `announceService` or `announceRepo` if things fail there. my bad.

orion/pkg/tests_integration/announce_integration_test.go
```diff
@@ -0,0 +1,140 @@
...
+	assert.EqualValues(t, 2, announces[1].Id)
+	assert.EqualValues(t, "Author 2", announces[1].Author)
+	assert.EqualValues(t, "Message 2", announces[1].Message)
+	assert.EqualValues(t, 3, announces[2].Id)
+	assert.EqualValues(t, "Author 3", announces[2].Author)
+	assert.EqualValues(t, "Message 3", announces[2].Message)
+	assert.EqualValues(t, 3, len(announces))
+}
+
+// It just occurred to me that this test makes no sense because unlike programs, announcements don't have a separate
```
ahsu1230: I agree. You can delete this test then.

orion/pkg/repos/announce_repo_test.go
```diff
@@ -0,0 +1,165 @@
...
+// Update
+//
+func TestUpdateAnnouncement(t *testing.T) {
+	db, mock, repo := initAnnounceTest(t)
+	defer db.Close()
+
+	// Mock DB statements and execute
+	now := time.Now().UTC()
+	result := sqlmock.NewResult(1, 1)
+	mock.ExpectPrepare("^UPDATE announcements SET (.*) WHERE id=?").ExpectExec().WithArgs(sqlmock.AnyArg(), now, "Author Name", "Valid Message", 1).WillReturnResult(result)
```
ahsu1230:
```
mock.ExpectPrepare("^UPDATE announcements SET (.*) WHERE id=?").
        ExpectExec().
        WithArgs(sqlmock.AnyArg(), now, "Author Name", "Valid Message", 1).
        WillReturnResult(result)
```

orion/pkg/repos/announce_repo_test.go
```diff
@@ -0,0 +1,165 @@
...
+// Create
+//
+func TestInsertAnnouncement(t *testing.T) {
+	db, mock, repo := initAnnounceTest(t)
+	defer db.Close()
+
+	// Mock DB statements and execute
+	now := time.Now().UTC()
+	result := sqlmock.NewResult(1, 1)
+	mock.ExpectPrepare("^INSERT INTO announcements").ExpectExec().WithArgs(sqlmock.AnyArg(), sqlmock.AnyArg(), now, "Author Name", "Valid Message").WillReturnResult(result)
```
ahsu1230: take a look at my comment below

orion/pkg/repos/announce_repo_test.go
```diff
@@ -0,0 +1,165 @@
...
+//
+// Select One
+//
+func TestSelectAnnouncement(t *testing.T) {
+	db, mock, repo := initAnnounceTest(t)
+	defer db.Close()
+
+	// Mock DB statements and execute
+	now := time.Now().UTC()
+	rows := sqlmock.NewRows([]string{"Id", "CreatedAt", "UpdatedAt", "DeletedAt", "PostedAt", "Author", "Message"}).AddRow(1, now, now, sql.NullTime{}, now, "Author Name", "Valid Message")
```
ahsu1230: take a look at my comment below for this statement and the one after this.

orion/pkg/repos/announce_repo_test.go
```diff
@@ -0,0 +1,165 @@
...
+// Test Select All
+//
+func TestSelectAllAnnouncements(t *testing.T) {
+	db, mock, repo := initAnnounceTest(t)
+	defer db.Close()
+
+	// Mock DB statements and execute
+	now := time.Now().UTC()
+	rows := sqlmock.NewRows([]string{"Id", "CreatedAt", "UpdatedAt", "DeletedAt", "PostedAt", "Author", "Message"}).AddRow(1, now, now, sql.NullTime{}, now, "Author Name", "Valid Message")
+	mock.ExpectPrepare("^SELECT (.+) FROM announcements").ExpectQuery().WillReturnRows(rows)
```
ahsu1230: apply comment from below to statements 31 & 32.

For the rows part, maybe it could look something like this:
```
rows := sqlmock.NewRows([]string{
    "Id",
    "CreatedAt",
    ....
})
```

ahsu1230: The format in `program_repo_test.go` should also follow this format too

PR #57, opened 2020-03-07
Added user repo, migrations, and domains

tonywu315:
@ahsu1230 
@mzhang2021


orion/pkg/domains/user.go
```diff
@@ -0,0 +1,69 @@
...
+var TABLE_USERS = "users"
+
+type User struct {
+	Id         uint
+	CreatedAt  time.Time    `db:"created_at"`
+	UpdatedAt  time.Time    `db:"updated_at"`
+	DeletedAt  sql.NullTime `db:"deleted_at"`
+	FirstName  string       `json:"firstName"`
+	LastName   string       `json:"lastName"`
+	MiddleName string       `json:"middleName"`
```
ahsu1230: this can be null, so change this to `sql.NullString`

orion/pkg/domains/user.go
```diff
@@ -0,0 +1,69 @@
...
+	Email      string       `json:"email"`
+	Phone      string       `json:"phone"`
+	IsGuardian bool         `json:"isGuardian"`
+	GuardianId uint         `db:"program_id" json:"guardianId"`
```
ahsu1230: This can also be nullable, so change this to `sql.NullUint`... But taking a look at `golang/sql`, there is no `sql.NullUint`.

Do you feel comfortable implementing our own `sql.NullUint`?
You can follow the following example from the Go open source.
https://github.com/golang/go/blob/5476967b1a3d29fef4061999c00cadbec19ac0e3/src/database/sql/sql.go#L184-L208
I would say don't put this NullUint struct in the repo package. Otherwise, we will have a circular dependency. We might want to create our own sql_helper package instead. And have this user.go file import that sql_helper package.

orion/pkg/domains/user.go
```diff
@@ -0,0 +1,69 @@
...
+	// Phone validation
+	if matches, _ := regexp.MatchString(REGEX_PHONE, phone); !matches || len(phone) > 24 {
+		return errors.New("invalid phone")
+	}
+
+	// Guardian Id validation
+	if guardianId == id {
+		return errors.New("invalid guardian id")
+	}
+
```
ahsu1230: Once you change `guardianId` to nullable, you can do an additional check. If `isGuardian` is true, there should not be a `guardianId` (unless we support grandparents, but probably not!).

orion/pkg/repos/user_repo.go
```diff
@@ -0,0 +1,182 @@
...
+			&user.IsGuardian,
+			&user.GuardianId); errScan != nil {
+			return results, errScan
+		}
+		results = append(results, user)
+	}
+	return results, nil
+}
+
+func (pr *userRepo) SelectById(id string) (domains.User, error) {
```
ahsu1230: `ur` for `*userRepo`
and id's here are integers not strings.

orion/pkg/repos/user_repo_test.go
```diff
@@ -0,0 +1,194 @@
...
+// Test Select All
+//
+func TestSelectAllUsers(t *testing.T) {
+	db, mock, repo := initUserTest(t)
+	defer db.Close()
+
+	// Mock DB statements and execute
+	now := time.Now().UTC()
+	rows := sqlmock.NewRows([]string{"Id", "CreatedAt", "UpdatedAt", "DeletedAt", "FirstName", "LastName", "MiddleName", "Email", "Phone", "IsGuardian", "GuardianId"}).
+		AddRow(1, now, now, sql.NullTime{}, "John", "Smith", "Middle", "john.smith@example.com", "555-555-0100", false, 0)
```
ahsu1230: since this table has a more columns, can you make this more "vertical"?
```
sqlmock.NewRows([]string{
   "Id",
   "CreatedAt",
   ...
}).
AddRow(
   1, 
    ...
)
```

orion/pkg/repos/user_repo_test.go
```diff
@@ -0,0 +1,194 @@
...
+// Select One
+//
+func TestSelectUser(t *testing.T) {
+	db, mock, repo := initUserTest(t)
+	defer db.Close()
+
+	// Mock DB statements and execute
+	now := time.Now().UTC()
+	rows := sqlmock.NewRows([]string{"Id", "CreatedAt", "UpdatedAt", "DeletedAt", "FirstName", "LastName", "MiddleName", "Email", "Phone", "IsGuardian", "GuardianId"}).
+		AddRow(1, now, now, sql.NullTime{}, "John", "Smith", "Middle", "john.smith@example.com", "555-555-0100", false, 0)
```
ahsu1230: same with these two

orion/pkg/sql_helper/sql.go
```diff
@@ -0,0 +1,27 @@
...
+	"database/sql/driver"
+)
+
+// NullUint represents an uint that may be null.
+// NullUint implements the Scanner interface so
+// it can be used as a scan destination, similar to NullString.
+type NullUint struct {
+	Uint  uint
+	Valid bool // Valid is true if Uint is not NULL
+}
```
ahsu1230: very nice!

orion/pkg/domains/user.go
```diff
@@ -0,0 +1,71 @@
...
+var TABLE_USERS = "users"
+
+type User struct {
+	Id         uint
+	CreatedAt  time.Time           `db:"created_at"`
+	UpdatedAt  time.Time           `db:"updated_at"`
+	DeletedAt  sql.NullTime        `db:"deleted_at"`
+	FirstName  string              `json:"firstName"`
+	LastName   string              `json:"lastName"`
+	MiddleName sql.NullString      `json:"middleName"`
```
ahsu1230: same thing as before.
Make `MiddleName` type string. Then in the Insert() and Upate() methods of the repo layer, give a sql.NullString instead of a string.

orion/pkg/domains/user.go
```diff
@@ -0,0 +1,71 @@
...
+	Email      string              `json:"email"`
+	Phone      string              `json:"phone"`
+	IsGuardian bool                `json:"isGuardian"`
+	GuardianId sql_helper.NullUint `db:"guardian_id" json:"guardianId"`
```
ahsu1230: For now turn this into a uint. And do the same as above. NullUint only in the Insert() and Update() layer of repo please.
Thank you!

PR #60, opened 2020-03-07
fixed announceEdit page

MiniViolin:


orion/sites/admin/src/announce/announce.js
```diff
@@ -2,12 +2,70 @@
...
+					<div className="announce-lists">
+						<li className="li-med"> Date </li>
+						<li className="li-med"> Author </li>
+						<li className="li-large"> Message </li>
+						<li className="li-small"> </li>
+					</div>
+				</ul>
+
+				{/* these are the lists under the subheading */}
+				<ul> {/*I think there's a problem having li under li, but I'll fix it later*/}
```
ahsu1230: You can remove this comment

orion/sites/admin/src/announce/announce.js
```diff
@@ -2,12 +2,70 @@
...
+		let fakeAnnounceC = {
+			date: "02/23/2020",
+			author: "Talkative",
+			message: "This will be a longer message later"
+		};
 		return (
       <div id="view-announce">
-        <h1>All Announcements</h1>
+        <h1>All Announcements ({numAnnouncements}) </h1>
+				<ul> {/* this is the subheading */}
```
ahsu1230: I think instead of leaving comments, we can try using "meaningful code" to express what we're trying to say in comments.
For example, this `<ul>` is a subheader.
So we can give it an id like so:
```
<ul id="subheader">
```
That way we don't need verbose comments :)

orion/sites/admin/src/announce/announce.js
ahsu1230: And instead, you can use `<ul id="announce-list">`

orion/sites/admin/src/announce/announce.js
```diff
@@ -2,12 +2,70 @@
...
+						<li className="li-med"> Date </li>
+						<li className="li-med"> Author </li>
+						<li className="li-large"> Message </li>
+						<li className="li-small"> </li>
+					</div>
+				</ul>
+
+				{/* these are the lists under the subheading */}
+				<ul> {/*I think there's a problem having li under li, but I'll fix it later*/}
+					<li className="announce-lists">
```
ahsu1230: Similarly here, `announce-lists` isn't the best name since technically this is not a list, it's an individual item of a list. So I suggest a different name like `announce-list-item` or `announce-list-row`. Whatever name makes sense for you. Still keep it a `className` since there are multiple of these `<li>`'s.

orion/sites/admin/src/announce/announce.js
```diff
@@ -2,12 +2,70 @@
...
+						<div className="li-med"> {fakeAnnounceC.date} </div>
+						<div className="li-med"> {fakeAnnounceC.author} </div>
+						<p className="li-large"> {fakeAnnounceC.message} </p>
+						<div className="li-small"> Edit > </div>
+					</li>
+
+				</ul>
+
+			{/* I think this should be <Link> not <a> but <Link> makes page not load. fix later*/}
```
ahsu1230: You can remove these comments that don't make sense

orion/sites/admin/src/announce/announce.js
```diff
@@ -2,12 +2,70 @@
 require('./announce.styl');
 import React from 'react';
 import ReactDOM from 'react-dom';
+import API from '../api.js';
+import { Modal } from '../modals/modal.js';
```
ahsu1230: You can remove API and Modal, they're not being used yet. Only include what you need.

orion/sites/admin/src/announce/announce.styl
```diff
@@ -5,9 +5,48 @@
   height: 400px
   padding: 0 spacing2x
   box-sizing: border-box
-  color: colorBlackDark
+  color: colorGrayDark
+  font-family: fontFamilyRoboto /*not working*/
```
ahsu1230: You sure this isn't working?

orion/sites/admin/src/announce/announce.styl
```diff
@@ -5,9 +5,48 @@
...
+  & ul
+    padding: 0
+
+  & .announce-lists
+    display: inline-block
+    list-style: none
+    padding-left: 0
+    height: lineHeight6
+    border-bottom: 1px solid colorGrayLight
+    line-height: lineHeight6
```
ahsu1230: When doing CSS properties, I would order like so:
```
display
height
padding-left
border-bottom
line-height
list-style
```

ahsu1230: Box Model stuff first (display, width, height, margin, padding, border)
Font stuff (colors, background-colors, text-align)
Then other things

orion/sites/admin/src/announce/announce.styl
```diff
@@ -5,9 +5,48 @@
...
+
+      & .li-med
+          width: spacing16x
+
+      & .li-large
+          width: spacing64x
+
+  .announcement-button
+    display: inline-block
+    background-color: colorTurquoise
```
ahsu1230: this line should be lower in the list of CSS properties

orion/sites/admin/src/announce/announceEdit.js
```diff
@@ -0,0 +1,65 @@
+'use strict';
+require('./announceEdit.styl');
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { Link } from 'react-router-dom';
+import API from '../api.js';
+import { Modal } from '../modals/modal.js';
+import { OkayModal } from '../modals/okayModal.js';
+import { YesNoModal } from '../modals/yesnoModal.js';
```
ahsu1230: Can remove all the API and Modal stuff now. Only include what you need.
This is important because if you're not vigilant about keeping code clean, you'll start to get a very messy codebase where you're not sure if things are still being used or not.

orion/sites/admin/src/announce/announceEdit.js
```diff
@@ -0,0 +1,65 @@
...
+
+export class AnnounceEditPage extends React.Component {
+    constructor(props) { /*this is just copied from programEdit, need to change sightly*/
```
ahsu1230: If you don't need this comment anymore, please remove it.

orion/sites/admin/src/announce/announceEdit.js
```diff
@@ -0,0 +1,65 @@
...
+      this.setState({[value]: event.target.value});
+    }
+
+      render() {
+/*I have no idea how this looks because this page has "Cannot GET /announceEdit.js" error. fix later*/
+        return (
+            <div id ="view-announce-edit">
+            <h2>Add Announcement</h2>
+
+            <h4>Date</h4> {/* will be post date later*/}
```
ahsu1230: I'm okay with leaving comments in the code, especially as notes for yourself.
Just make sure it's accurate and more clear.

So for example, here you can make a more "actionable" comment.
```
{/* TODO: Make this "Posted Date" */}
```

orion/sites/admin/src/announce/announceEdit.js
```diff
@@ -0,0 +1,65 @@
...
+      //this.onSaved = this.onSaved.bind(this);
+      //this.onDismissModal = this.onDismissModal.bind(this);
+    }
+
+    handleChange(event, value) {
+      this.setState({[value]: event.target.value});
+    }
+
+      render() {
+/*I have no idea how this looks because this page has "Cannot GET /announceEdit.js" error. fix later*/
```
ahsu1230: If you click on the "Add button", you should be able to see this page right? Edit won't work yet, but Add should.

orion/sites/admin/src/index.js
```diff
@@ -15,13 +15,16 @@ import { ProgramEditPage } from './programs/programEdit.js';
...
+import { AnnounceEditPage } from './announce/announceEdit.js';
 import { LocationPage } from './location/location.js';
 import { LocationEditPage } from './location/locationEdit.js';
 
 const Achieve = () => <AchievePage/>;
 const AchieveEdit = () => <AchieveEditPage/>;
 const AchieveEditMatch = ({match}) => <AchieveEditPage year={match.params.year}/>;
 const Announce = () => <AnnouncePage/>;
+const AnnounceEdit = () => <AnnounceEditPage/>;
+const AnnounceEditMatch = ({match}) => <AnnounceEditPage programId={match.params.AnnounceId}/>;
```
ahsu1230: looks like you put `programId=...` should probably be something else

orion/sites/admin/src/index.js
```diff
@@ -51,6 +54,8 @@ class App extends React.Component {
           <Route path="/program/:programId/edit" component={ProgramEditMatch}/>
           <Route path="/programs/add" component={ProgramEdit}/>
           <Route path="/programs" component={Programs}/>
+          <Route path="/announce/:announceId/edit" component={AnnounceEditMatch}/>
```
ahsu1230: nice!

orion/sites/admin/src/index.js
ahsu1230: the `params` part should match with the variable name below on line 57.
So if the path is `"/announce/:announceId/edit"`, the variable is `announceId`. Lower-cased not Upper-cased.

orion/sites/admin/src/index.js
```diff
@@ -51,6 +54,8 @@ class App extends React.Component {
...
+          <Route path="/announce/add" component={AnnounceEdit}/>
```
ahsu1230: Ah another thought, below we have `"/announcements"` as a path. Since these all do with announcements, let's make them consistent. For the other paths, can you make them:
```
"/announcements/:announceId/edit"
"/announcements/add"
```

Thanks!

orion/sites/admin/src/announce/announceEdit.js
MiniViolin: Yes, it should work now, I just forgot to remove the comment.

orion/sites/admin/src/announce/announceEdit.styl
```diff
@@ -0,0 +1,49 @@
...
+    line-height: lineHeight0
+    font-size: fontSize0
+
+  & .buttons
+    margin-top: spacing2x
+    margin-bottom: spacing8x
+
+    & button
+      padding: spacing2x spacing4x
+      font-size: fontSize0
```
ahsu1230: move this towards the bottom (but before cursor)

ahsu1230:
nice job! Looks good to merge!

PR #61, opened 2020-03-07
Locations edit

mythicat554:


ahsu1230:
@mythicat554 Ah in case you're not sure, I've approved this PR. lgtm! 👍

PR #72, opened 2020-03-21
User integration

tonywu315:
@ahsu1230 
@mzhang2021 
This also uses code in #57


orion/pkg/repos/migrations/000007_create_table_users.up.sql
```diff
@@ -9,8 +9,8 @@ CREATE TABLE users
     middle_name varchar(32),
     email       varchar(64)  NOT NULL,
     phone       varchar(24)  NOT NULL,
-    isGuardian  boolean      NOT NULL DEFAULT 0,
-    guardianId  int unsigned,
+    is_guardian boolean      NOT NULL DEFAULT 0,
+    guardian_id int unsigned,
```
ahsu1230: if you end up changing the migration file, your current sql might have already created the user table with `isGuardian` and `guardianId` columns. To fix this, you will have to fix your migration state. Use these commands:
```
migrate -source file://pkg/repos/migrations -database mysql://USER:PASSWORD@\(localhost\)/mathnavdb down 1
migrate -source file://pkg/repos/migrations -database mysql://USER:PASSWORD@\(localhost\)/mathnavdb up 1
```
This will move your migration from version 7 to 6 (drops the user table). Then it will move you back to 7 (create user table again).

orion/pkg/controllers/user_controller_test.go
```diff
@@ -282,16 +282,16 @@ func createMockUser(id uint, firstName, lastName, middleName, email, phone strin
...
 		Phone:      phone,
 		IsGuardian: isGuardian,
-		GuardianId: guardianId,
+		GuardianId: domains.CreateNullUint(guardianId),
 	}
 }
 
 func createBodyFromUser(user domains.User) io.Reader {
-	marshal, err := json.Marshal(user)
+	marshal, err := json.Marshal(&user)
```
ahsu1230: Overall, I think everything looks decently easy to use. It's just this part that kind of scares me because it's so nuanced. When marshaling json, especially a struct with nested structs inside it that has custom Marshaling, you must give it a pointer. `json.Marshal(...)` takes an `interface{}` typed parameter and for some reason (I found a source before, but can't find it atm) it won't call the custom marshaller unless you pass the pointer of the struct instead.
One thing we could do is to make sure we ALWAYS pass in a pointer of a struct to `json.Marshal` so we never bump into this nuanced problem. But yeah - it's nuanced and errors that originate from not including `&` can be misleading at times. :(

orion/pkg/domains/alias_domains.go
```diff
@@ -0,0 +1,79 @@
...
+func CreateNullUint(num uint) NullUint {
+	return NullUint{
+		Num:   num,
+		Valid: num != 0,
+	}
+}
+
+// Scan implements the Scanner interface.
+func (n *NullUint) Scan(value interface{}) error {
+	num, ok := value.(int64)
```
ahsu1230: This is an golang/sql thing. You can only cast `interface{}` as `int64` or one of the very few supported value types. https://stackoverflow.com/questions/39379928/golang-save-enum-type-to-sql-database-panic-sql-converting-exec-argument-1

Once you do, you can cast it normally however you'd like.

orion/pkg/sql_helper/sql.go
```diff
@@ -1,27 +0,0 @@
-package sql_helper
```
ahsu1230: moved to alias_domains

orion/pkg/tests_integration/user_integration_test.go
```diff
@@ -0,0 +1,180 @@
...
+	user3 := createUser(
+		"Foo",
+		"Bar",
+		"Middle3",
+		"foobar@example.com",
+		"555-555-0102",
+		false,
+		1,
+	)
+	body1 := createJsonBody(&user1)
```
ahsu1230: Again note how we pass the pointers of these structs and not the structs themselves when Json Marshaling

orion/pkg/domains/user.go
```diff
@@ -13,14 +13,14 @@ type User struct {
 	Id         uint
 	CreatedAt  time.Time    `db:"created_at"`
 	UpdatedAt  time.Time    `db:"updated_at"`
-	DeletedAt  sql.NullTime `db:"deleted_at"`
+	DeletedAt  sql.NullTime `db:"deleted_at" json:"-"`
```
ahsu1230: this part will signal to not include this field in the resulting JSON. So web clients will not be able to see `DeletedAt`

orion/pkg/domains/user_valid_test.go
```diff
@@ -9,7 +9,7 @@ import (
 
 func TestValidFirstName(t *testing.T) {
 	// Checks for valid first names
-	user := domains.User{FirstName: "John", LastName: "Smith", Email: "gmail@gmail.com", Phone: "555-555-0100", IsGuardian: true}
+	user := domains.User{FirstName: "John", LastName: "Smith", Email: "gmail@gmail.com", Phone: "555-555-0100", IsGuardian: true, GuardianId: domains.NewNullUint(0)}
```
ahsu1230: nit: I think this line is too long.
Can you separate the lines into a block?
domain.User{
FirstName: ...,
...
}

ahsu1230:
small nit. feel free to merge whenever you're happy with it.

PR #69, opened 2020-03-20
Class migration domain repo

tonywu315:
@ahsu1230 
@mzhang2021

PR #73, opened 2020-03-21
Class controller

tonywu315:
@ahsu1230 
@mzhang2021 
This uses code in #69


orion/pkg/controllers/class_controller_test.go
```diff
@@ -0,0 +1,539 @@
...
+
+	// Create new HTTP request to endpoint
+	recorder := sendHttpRequest(t, http.MethodGet, "/api/classes/v1/class/program2_2020_summer", nil)
+
+	// Validate results
+	assert.EqualValues(t, http.StatusNotFound, recorder.Code)
+}
+
+//
+// Test Get Classes By Program and Semester
```
ahsu1230: You can just make this generic
// Test Get Classes by other properties

orion/pkg/controllers/class_controller_test.go
```diff
@@ -0,0 +1,539 @@
...
+			{
+				Id:         1,
+				ProgramId:  "program1",
+				SemesterId: "2020_spring",
+				ClassKey:   sql.NullString{String: "class1", Valid: true},
+				ClassId:    "program1_2020_spring_class1",
+				LocationId: "churchill",
+				Times:      "3 pm - 5 pm",
+				StartDate:  now,
+				EndDate:    now.Add(time.Hour * 24 * 60),
```
ahsu1230: You can just create two variables:
```
now := time.Now().UTC()
later := now.Add(...)
```
And just use these two variables instead of having to keep re-calculating the EndDate

orion/pkg/router/handler.go
```diff
@@ -22,7 +22,17 @@ func (h *Handler) SetupApiEndpoints() {
...
+	apiClasses := h.Engine.Group("api/classes/")
+	{
+		apiClasses.GET("/v1/all", controllers.GetAllClasses)
+		apiClasses.POST("/v1/create", controllers.CreateClass)
+		apiClasses.GET("/v1/class/:classId", controllers.GetClassById)
+		apiClasses.GET("/v1/classes/program/:programId", controllers.GetClassesByProgram)
+		apiClasses.GET("/v1/classes/semester/:semesterId", controllers.GetClassesBySemester)
+		apiClasses.GET("/v1/classes/program/:programId/semester/:semesterId", controllers.GetClassesByProgramAndSemester)
+		apiClasses.POST("/v1/class/:classId", controllers.UpdateClass)
+		apiClasses.DELETE("/v1/class/:classId", controllers.DeleteClass)
```
ahsu1230: Move this endpoint and the UpdateClass endpoint to right below the GetClassById.
We want to keep similar api endpoint forms (`v1/class/:classId`) close to each other.
And then sort by Get -> Post -> Deletes.

orion/pkg/controllers/class_controller_test.go
```diff
@@ -0,0 +1,555 @@
...
+				Id:         3,
+				ProgramId:  "program1",
+				SemesterId: "2020_fall",
+				ClassKey:   "class2",
+				ClassId:    "program1_2020_fall_class2",
+				LocationId: "churchill",
+				Times:      "5 pm - 7 pm",
+				StartDate:  now,
+				EndDate:    later3,
+			},
```
ahsu1230: consider making this a global variable or from a `createFakeClasses()` helper function which you can then use for these 3-4 similar tests (maybe even for all the tests if you'd like).
Make the mock function return the result from `createFakeClasses()` or global variable and do the rest of the testing in each unit tests!

orion/pkg/controllers/class_controller_test.go
```diff
@@ -0,0 +1,445 @@
...
+	}
+
+	assertMockClasses(t, 3, classes[0])
+	assertMockClasses(t, 4, classes[1])
+	assert.EqualValues(t, 2, len(classes))
+}
+
+func TestGetClassesByProgramAndSemester_Success(t *testing.T) {
+	classService.mockGetByProgramAndSemesterId = func(programId, semesterId string) ([]domains.Class, error) {
+		return createMockClasses(), nil
```
ahsu1230: I think this is the culprit to your tests failing. You are mocking the service to return ALL mocked classes. So the list of "found" programs and semester becomes the entire mocked class list. Consider filtering out classes here.

PR #74, opened 2020-03-21
sessions backend

mzhang2021:
@ahsu1230 
@tonywu315


ahsu1230:
I believe the Classes database repo is in now, so you can actually start testing to see if sessions work. The only thing is, I don't think we'll ever use a getAllSessions() endpoint. Instead of change it to getAllSessionsByClassId().

This means given a classId which is a string, return all sessions with associated classId (foreign key). However, we will still use getAllSessionById() as well which should return one session (given its row id).

orion/pkg/repos/migrations/000006_create_table_sessions.up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE sessions (
+	id int unsigned NOT NULL AUTO_INCREMENT,
+	created_at datetime NOT NULL,
+	updated_at datetime NOT NULL,
+	deleted_at datetime,
+	class_id varchar(64) NOT NULL,
```
ahsu1230: double check that `class_id` varchar length is the same as in the classes table.

orion/pkg/repos/migrations/000006_create_table_sessions.up.sql
```diff
@@ -0,0 +1,13 @@
...
+	starts_at datetime NOT NULL,
+	ends_at datetime NOT NULL,
+	canceled boolean NOT NULL DEFAULT false,
```
ahsu1230: I believe in MySQL the default of a boolean is 0. false will work, but i think it is actually 0.

orion/pkg/repos/migrations/000006_create_table_sessions.up.sql
```diff
@@ -0,0 +1,13 @@
...
+	notes text NOT NULL,
+	PRIMARY KEY (id),
+	FOREIGN KEY (class_id) REFERENCES classes(class_id)
```
ahsu1230: nice!

orion/pkg/tests_integration/session_integration_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/ahsu1230/mathnavigatorSite/orion/pkg/domains"
+	"github.com/stretchr/testify/assert"
+	"net/http"
+	"testing"
+	"time"
+)
+
+// Test: Create 3 Sessions, 2 With Same Class Id, and GetAllByClassId()
+func Test_CreateSessions(t *testing.T) {
+	resetTable(t, domains.TABLE_SESSIONS)
```
ahsu1230: I think we'll also need to resetTable for Classes too. I think this needs to be called after resetting table sessions due to the foreign key constraint. Because sessions depend on classes existing, we can't remove classes first because the sessions table will complain it doesn't have correctly mapped classIds.

ahsu1230:
Cool, it's looking good. I think Tony is not done with the classes controllers yet so that's why integration tests won't work yet.

orion/pkg/tests_integration/session_integration_test.go
```diff
@@ -0,0 +1,183 @@
...
+	assert.EqualValues(t, "cancelled due to corona", session.Notes)
+}
+
+// Test: Create 1 Session, Delete it, GetBySessionId()
+func Test_DeleteSession(t *testing.T) {
+	resetTable(t, domains.TABLE_SESSIONS)
+	resetTable(t, domains.TABLE_CLASSES)
+	resetTable(t, domains.TABLE_SEMESTERS)
+	resetTable(t, domains.TABLE_LOCATIONS)
+	resetTable(t, domains.TABLE_PROGRAMS)
```
ahsu1230: Nice! lol consider putting this into a helper function for this class? You repeat this sequence of lines a few times in this file

orion/pkg/tests_integration/session_integration_test.go
```diff
@@ -0,0 +1,183 @@
...
+	assert.EqualValues(t, http.StatusOK, recorder1.Code)
+	assert.EqualValues(t, http.StatusOK, recorder2.Code)
+	assert.EqualValues(t, http.StatusOK, recorder3.Code)
+	assert.EqualValues(t, http.StatusOK, recorder4.Code)
+	assert.EqualValues(t, http.StatusOK, recorder5.Code)
+	assert.EqualValues(t, http.StatusOK, recorder6.Code)
+	assert.EqualValues(t, http.StatusOK, recorder7.Code)
+	assert.EqualValues(t, http.StatusOK, recorder8.Code)
+	assert.EqualValues(t, http.StatusOK, recorder9.Code)
+	assert.EqualValues(t, http.StatusOK, recorder10.Code)
```
ahsu1230: lol I guess all of this from line 24 is kind of necessary. I wonder if we can make it better

orion/pkg/domains/session.go
```diff
@@ -0,0 +1,40 @@
...
+	Notes     string       `json:"notes"`
+}
+
+func (session *Session) Validate() error {
+	// Retrieves the inputted values
+	classId := session.ClassId
+	notes := session.Notes
+
+	// Class ID validation
+	if matches, _ := regexp.MatchString(REGEX_GENERIC_ID, classId); !matches {
```
ahsu1230: I think this is unneeded because if classId is not valid, the database will scream because it's not an existing classId. So I don't think we have to check classId here

orion/pkg/repos/migrations/000007_create_table_sessions.up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE sessions (
+	id int unsigned NOT NULL AUTO_INCREMENT,
+	created_at datetime NOT NULL,
+	updated_at datetime NOT NULL,
+	deleted_at datetime,
+	class_id varchar(192) NOT NULL,
+	starts_at datetime NOT NULL,
+	ends_at datetime NOT NULL,
+	canceled boolean NOT NULL DEFAULT 0,
+	notes text NOT NULL,
```
ahsu1230: ah sorry forgot to mention. notes here can actually be NULLABLE. `notes` are little blurbs that the admin can use to display to students like "Room change today to 124!"

Just make this Nullable here, but don't worry about the whole NullString debacle yet. Wait for #72 to merge, which will make it easier to use NullStrings in places.

orion/pkg/router/handler.go
```diff
@@ -29,13 +29,14 @@ func (h *Handler) SetupApiEndpoints() {
 		apiClasses.GET("/v1/class/:classId", controllers.GetClassById)
 		apiClasses.POST("/v1/class/:classId", controllers.UpdateClass)
 		apiClasses.DELETE("/v1/class/:classId", controllers.DeleteClass)
+		apiClasses.GET("/v1/class/:classId/sessions", controllers.GetAllSessionsByClassId)
```
ahsu1230: Sorry again. I looked at this again and thought it was confusing.
Can you change this back to `/v1/sessions/class/:classId`
It should represent "getting a bunch of sessions with the class `:classId`"

And place it after the DELETE sessions endpoint. We want to keep the X by Id endpoints together for readability.

orion/pkg/repos/migrations/000007_create_table_sessions.up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE sessions (
```
ahsu1230: If the users sql files are already `00007` you're going to have to rename your files `00008`. This is because if you change the ordering, everybody's migrations will be messed up. Always append!

orion/pkg/controllers/session_controller.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/gin-gonic/gin"
+	"net/http"
+)
+
+func GetAllSessionsByClassId(c *gin.Context) {
+	classId := c.Param("classId")
+
+	sessionList, err := services.SessionService.GetAllByClassId(classId)
+	if err != nil {
+		c.String(http.StatusInternalServerError, err.Error())
```
ahsu1230: From a previous PR that I made last week. Make sure that before you use c.String(...), use c.Error(err) so the error will be displayed on the webserver side.

orion/pkg/controllers/session_controller.go
```diff
@@ -0,0 +1,85 @@
...
+	if err := sessionJson.Validate(); err != nil {
+		c.String(http.StatusBadRequest, err.Error())
+		return
+	}
+
+	err := services.SessionService.Create(sessionJson)
+	if err != nil {
+		c.String(http.StatusInternalServerError, err.Error())
+	} else {
+		c.JSON(http.StatusOK, nil)
```
ahsu1230: Also from last week's PR, if you are not returning a body, just use c.Status(OK)

mzhang2021:
I changed the api endpoint, but that ended up invoking the following error message: "handlers are already registered for path '/api/classes/v1/class/:classId'"
I suspect this has something to do with [this](https://github.com/gin-gonic/gin/issues/196), but my path is different ('/api/sessions/v1/class/:classId'), so I'm not sure why this is happening.

tonywu315:
> I changed the api endpoint, but that ended up invoking the following error message: "handlers are already registered for path '/api/classes/v1/class/:classId'"
> I suspect this has something to do with [this](https://github.com/gin-gonic/gin/issues/196), but my path is different ('/api/sessions/v1/class/:classId'), so I'm not sure why this is happening.

In handler.go, you used `apiClasses.GET("/v1/class/:classId",` controllers.GetAllSessionsByClassId)` which is the same endpoint as controllers.GetClassById

orion/pkg/router/handler.go
```diff
@@ -72,6 +71,7 @@ func (h *Handler) SetupApiEndpoints() {
 		apiSessions.GET("/v1/session/:id", controllers.GetSessionById)
 		apiSessions.POST("/v1/session/:id", controllers.UpdateSession)
 		apiSessions.DELETE("/v1/session/:id", controllers.DeleteSession)
+		apiClasses.GET("/v1/class/:classId", controllers.GetAllSessionsByClassId)
```
ahsu1230: I think I meant in this line `/v1/sessions/class/:classId`

ahsu1230: OH nvm I see it. You're using apiClasses here

mzhang2021:
Right, but the thing is under the "h.Engine.Group("api/sessions/")" block, so the path should be different as '/api/sessions/v1/class/:classId', no?

orion/pkg/router/handler.go
mzhang2021: Oh I'm silly. My bad.

orion/pkg/controllers/session_controller.go
ahsu1230: *bump* please use c.Status

orion/pkg/controllers/session_controller.go
```diff
@@ -0,0 +1,92 @@
...
+		c.String(http.StatusBadRequest, err.Error())
+		return
+	}
+
+	err := services.SessionService.Update(id, sessionJson)
+	if err != nil {
+		c.Error(err)
+		c.String(http.StatusInternalServerError, err.Error())
+	} else {
+		c.JSON(http.StatusOK, nil)
```
ahsu1230: c.Status()

orion/pkg/controllers/session_controller_test.go
```diff
@@ -0,0 +1,229 @@
...
+	now := time.Now().UTC()
+	sessionService.mockGetAllByClassId = func(classId string) ([]domains.Session, error) {
+		return []domains.Session{
+			{
+				Id:       1,
+				ClassId:  "id_1",
+				StartsAt: now,
+				EndsAt:   now,
+				Canceled: false,
+				Notes:    domains.NewNullString("special lecture from guest"),
```
ahsu1230: ooh nice!!

orion/pkg/domains/session.go
```diff
@@ -0,0 +1,36 @@
...
+type Session struct {
+	Id        uint
+	CreatedAt time.Time    `db:"created_at"`
+	UpdatedAt time.Time    `db:"updated_at"`
+	DeletedAt sql.NullTime `db:"deleted_at"`
+	ClassId   string       `db:"class_id" json:"classId"`
+	StartsAt  time.Time    `db:"starts_at" json:"startsAt"`
+	EndsAt    time.Time    `db:"ends_at" json:"endsAt"`
+	Canceled  bool         `db:"canceled" json:"canceled"`
+	Notes     NullString   `db:"notes" json:"notes"`
```
ahsu1230: awesome possums! Hopefully this just makes all our lives so much easier!

PR #75, opened 2020-03-21
Class integration

tonywu315:
@ahsu1230 
@mzhang2021 
This uses code from #69


orion/pkg/tests_integration/class_integration_test.go
```diff
@@ -0,0 +1,493 @@
...
+	"testing"
+	"time"
+)
+
+// Test: Create 3 Classes and GetAll()
+func Test_CreateClasses(t *testing.T) {
+	resetTable(t, domains.TABLE_CLASSES)
+
+	now := time.Now().UTC()
+	class1 := createClass(
```
ahsu1230: I think this may not work because in order to create a class, you must first create a program, a semester, and a location in order for the foreign keys to work.
In the unit tests, you didn't have to do any of this because it was all mocked. With an integration test, there's a real MySQL DB working in the background.

So looks like you'll probably need helper function(s) to create a program, a semester, and a location.

orion/pkg/tests_integration/class_integration_test.go
```diff
@@ -0,0 +1,493 @@
...
+		"program1",
+		"2020_spring",
+		sql.NullString{String: "class1", Valid: true},
+		"program1_2020_spring_class1",
+		"churchill",
+		"3 pm - 5 pm",
+		now,
+		now.Add(time.Hour*24*60),
```
ahsu1230: Can use variables to define these endDates and simply use those variables instead of having to re-calculate them.

orion/pkg/tests_integration/class_integration_test.go
```diff
@@ -0,0 +1,284 @@
...
+// Helper methods
+func createClass(id int) domains.Class {
+	now := time.Now().UTC()
+	later1 := now.Add(time.Hour * 24 * 30)
+	later2 := now.Add(time.Hour * 24 * 31)
+	later3 := now.Add(time.Hour * 24 * 60)
+	switch id {
+	case 1:
+		return domains.Class{
+			ProgramId:  "program1",
```
ahsu1230: View my comment before about needing to create a Programs object, Semesters object and locations object before persisting trying to insert a class into db.

Basically here you're trying to insert a class linked to a program with id `program1`. But there hasn't been a program even inserted into the db yet! So that's why these creates are giving you errors.
In addition, every time you reset the tables for the test, you may want to consider clearing the other tables as well.

orion/pkg/tests_integration/class_integration_test.go
```diff
@@ -0,0 +1,337 @@
...
+	assertClass(t, 4, classes[1])
+	assert.EqualValues(t, 2, len(classes))
+}
+
+// Test: Create 4 Classes and GetClassesByProgramAndSemester()
+func Test_GetClassesByProgramAndSemester(t *testing.T) {
+	resetTable(t, domains.TABLE_CLASSES)
+	resetTable(t, domains.TABLE_PROGRAMS)
+	resetTable(t, domains.TABLE_SEMESTERS)
+	resetTable(t, domains.TABLE_LOCATIONS)
```
ahsu1230: nit: maybe use a helper function for these resets too?

PR #76, opened 2020-03-25
Get announcements sorted

mzhang2021:
@ahsu1230 
@tonywu315


ahsu1230:
can we get tests for this? I think one in repo and one as an integration would be good.

PR #77, opened 2020-03-25
Get achievements by year

mzhang2021:
@ahsu1230 
@tonywu315


ahsu1230:
Looking at this JSON body:
```
[
    { 
        year: 2018, 
        achievements: [ {achieveObj}, {achieveObj} ] 
    },
    { 
        year: 2017, 
        achievements: [ {achieveObj}, {achieveObj} ] 
    },
    ...
]
```

To me, this looks like (on the root level), to be a list of objects/structs.
Inside each struct, there's a `year` field (number) and an `achievements` field which is a list of structs. The inner list (achievements field) is a list of the `Achievement` object that's already in the codebase.

So I think in order to do this task, we may need to define our own struct (that contains year and achievements). Maybe we're also going to need to use Golang maps? So maybe we can somehow use maps going from year -> list of achievements?
At least try that out first. First use the SelectAll to get a list of all achievements, create this map, and get our final list of structs.

After that, there is a SQL GROUP BY that can help us with the first part. Straight from the database, I thiiiinkk we can retrieve a map of year -> list of achievements. Maybe some stack overflow research is in order!

orion/pkg/controllers/achieve_controller.go
```diff
@@ -28,6 +28,15 @@ func GetAchievementById(c *gin.Context) {
 	}
 }
 
+func GetAchievementsByYear(c *gin.Context) {
+	achieveList, err := services.AchieveService.GetByYear()
+	if err != nil {
+		c.String(http.StatusInternalServerError, err.Error())
```
ahsu1230: c.Error(err) before c.String

orion/pkg/domains/achieve.go
```diff
@@ -18,6 +18,11 @@ type Achieve struct {
 	Message   string       `json:"message"`
 }
 
+type AchieveList struct {
```
ahsu1230: anyway we can rename this to represent some sort of grouping with years? Like `AchieveYearList` or `AchieveYearGroup` or `AchieveByYear`?

ahsu1230: The reason why I'm whining about this is because `AchieveList` is a bit too general and doesn't really convey what this struct represents.

Imagine a new intern coming (which may very well soon come) and looking at this struct, they'd probably have to guess what this does or ask the originator (you) what this does. And if this cycle continues, in which a lot of interns come and go, you can probably guess how quickly this codebase can erode.

orion/pkg/domains/achieve.go
```diff
@@ -18,6 +18,11 @@ type Achieve struct {
 	Message   string       `json:"message"`
 }
 
+type AchieveYearGroup struct {
+	Year         uint
```
ahsu1230: ah could you please include struct tags here `json:"year"` and for achievements below.
I want to make sure when these make it to the FE, they're lowercased.

orion/pkg/router/handler.go
```diff
@@ -56,6 +56,7 @@ func (h *Handler) SetupApiEndpoints() {
 		apiAchieves.GET("/v1/achievement/:id", controllers.GetAchievementById)
 		apiAchieves.POST("/v1/achievement/:id", controllers.UpdateAchievement)
 		apiAchieves.DELETE("/v1/achievement/:id", controllers.DeleteAchievement)
+		apiAchieves.GET("/v1/achievements/years", controllers.GetAllAchievementsGroupedByYear)
```
ahsu1230: can just be "/v1/years"

PR #80, opened 2020-03-28
automated lists

MiniViolin:


orion/sites/admin/src/announce/announce.js
```diff
@@ -13,43 +15,27 @@ export class AnnouncePage extends React.Component {
...
-			message: "Watermelons are fruit, therefore they deserve to dry on the vine"
-		};
-		/* TODO: will use this to test text wrapping later */
-		let fakeAnnounceC = {
-			date: "02/23/2020",
-			author: "Talkative",
-			message: "This will be a longer message later"
-		};
+		const rows = this.state.list.map ((row, index) => {
+	      return <AnnounceRow key = {index} row = {row}/>
```
ahsu1230: nit:
just make this `key={index}` and `row={row}` (without the spaces in between). The idea is to keep the parameter groups together. So it should look like:
```
<AnnounceRow key={index} row={row}/>
```

orion/sites/admin/src/announce/announce.js
```diff
@@ -13,43 +15,27 @@ export class AnnouncePage extends React.Component {
...
-
-				<ul>
-					<AnnounceRow announceObj = {fakeAnnounceA}/>
-					<AnnounceRow announceObj = {fakeAnnounceB}/>
-					<AnnounceRow announceObj = {fakeAnnounceC}/>
-				</ul>
-				<Link to={"/announcements/add"}> <button className="announcement-button"> Add Announcement</button> </Link>
-			</div>
-		);
+					<ul id="view-announce">
```
ahsu1230: I don't think the id here should also be called `view-announce`. `view-announce` is supposed to be the styles for the overall page in general. We're trying to style a list here.

ahsu1230:
I believe we have to merge #67 first before merging this one.

ahsu1230:
lgtm!! Go ahead and merge this one!

PR #86, opened 2020-03-30
Hookup Announcements with API + postedAt DateTime

ahsu1230:
@MiniViolin @mythicat554 

Please take a look sometime and review my code. If there's anything that looks weird or you might have a question, do post something! Don't hold back!
This change does the following:

- Hooks up the announcement page with API & Modals
- Moves around styling to display the rows in a neat way.
- Introduces a postedAt field that will assumes we are posting immediately and successfully sends to backend to be saved.

- In the near future, we will use a DateTime picker to allow a user to easily select what day and what time to post their announcement.
- We could consider a front-end utility class to help us with DateTime stuff (since the Javascript DateTime has so many options).

If it all looks good, please approve this PR. Thanks!


orion/pkg/domains/announce.go
```diff
@@ -14,7 +14,7 @@ type Announce struct {
 	CreatedAt time.Time    `db:"created_at"`
 	UpdatedAt time.Time    `db:"updated_at"`
 	DeletedAt sql.NullTime `db:"deleted_at"`
-	PostedAt  time.Time    `db:"posted_at"`
+	PostedAt  time.Time    `db:"posted_at" json:"postedAt"`
```
ahsu1230: backend change

orion/sites/admin/src/static.styl
```diff
@@ -46,6 +46,9 @@ spacing20x = (spacing1x * 20)
 spacing24x = (spacing1x * 24)
 spacing30x = (spacing1x * 30)
 spacing32x = (spacing1x * 32)
+spacing36x = (spacing1x * 36)
+spacing42x = (spacing1x * 42)
+spacing48x = (spacing1x * 48)
```
ahsu1230: sometimes I wonder if all these are necessary.

orion/sites/admin/src/announce/announceEdit.js
```diff
@@ -55,3 +153,8 @@ export class AnnounceEditPage extends React.Component {
     );
   }
 }
+
+function createDate() {
+  let now = new Date(Date.now());
+  return now;
```
ahsu1230: add a TODO. We are going to use a DateTime selector to figure out how to select a certain day and at what time to post an announcement.

MiniViolin:
This looks pretty good to me, I don't see anything strange.

PR #91, opened 2020-03-30
Hide json

tonywu315:
@ahsu1230 
@mzhang2021


orion/pkg/domains/session.go
```diff
@@ -13,12 +13,12 @@ type Session struct {
 	Id        uint
 	CreatedAt time.Time    `db:"created_at"`
 	UpdatedAt time.Time    `db:"updated_at"`
-	DeletedAt sql.NullTime `db:"deleted_at"`
+	DeletedAt sql.NullTime `db:"deleted_at" json:"-"`
 	ClassId   string       `db:"class_id" json:"classId"`
 	StartsAt  time.Time    `db:"starts_at" json:"startsAt"`
 	EndsAt    time.Time    `db:"ends_at" json:"endsAt"`
-	Canceled  bool         `db:"canceled" json:"canceled"`
-	Notes     NullString   `db:"notes" json:"notes"`
```
ahsu1230: In this case I actually prefer to be explicit here.
I would prefer to know that the Canceled field will turn into `canceled` and `canceled` for both db and json. Guessing what things may turn out to be is kind of a pain in the butt.

ahsu1230:
You know what, let's start making our struct db/json tags as consistent as possible now.
Could you please extend this change to make the following changes:

- also hide away CreatedAt and UpdatedAt from json
- make all `Id` to to lowercase json `id` <- this may momentarily break some of our front-end pages.
- all fields should always explicitly declare their `json` counterpart.

```
Id        uint                      `json:"id"`
CreatedAt time.Time     `json:"-" db:"created_at"`
UpdatedAt time.Time    `json:"-" db:"updated_at"`
DeletedAt sql.NullTime `json:"-" db:"deleted_at"`
Year      uint                    `json:"year"`
Message   string            `json:"message"`
```

Technically, we don't need a `db:"message"` since go will

orion/pkg/domains/session.go
tonywu315: So which fields do not need a db tag?

ahsu1230: oh sorry. This is a confusing comment. Ignore what I said. You're right here. "db" tag is not needed because SQL is case-insensitive. So Canceled & canceled are the same thing.

PR #93, opened 2020-03-31
added feature attribute to program

mzhang2021:
@ahsu1230 
@tonywu315


mzhang2021:
Wasn't sure if I should do validation for featured since the range of possible featured values is subject to change.

orion/pkg/repos/migrations/000001_create_table_programs.up.sql
```diff
@@ -9,6 +9,7 @@ CREATE TABLE programs
     grade1      tinyint unsigned NOT NULL,
     grade2      tinyint unsigned NOT NULL,
     description text             NOT NULL,
+    featured	tinyint unsigned NOT NULL DEFAULT 0,
```
ahsu1230, line 12: can we create a new migration file for this? I think everybody is on database migration versions 7-8-9. If we add this feature on version 1, it means in order for it to work, everybody would have to blow out their database (remove all tables and start over). Create a new migration file that ALTERS database table programs and ADDs the featured column. Similarly the `.down.sql` file should just remove the column.

It's also a good practice with the idea of database migrations :)

mzhang2021:
For the down file, I tried to implement something to the effect of "DROP COLUMN IF EXISTS featured", but that doesn't seem to be an option. The stackoverflow thread below offers a workaround, but it doesn't seem worth it:
https://stackoverflow.com/questions/173814/using-alter-to-drop-a-column-if-it-exists-in-mysql/173820

PR #96, opened 2020-03-31
Get user

tonywu315:
@ahsu1230 
@mzhang2021


orion/pkg/repos/user_repo.go
```diff
@@ -89,6 +90,41 @@ func (ur *userRepo) SelectById(id uint) (domains.User, error) {
 	return user, errScan
 }
 
+func (ur *userRepo) SelectByGuardianId(guardianId domains.NullUint) ([]domains.User, error) {
```
ahsu1230, line 93: I think it's just here where I have an issue.
If the signature of this function is `NullUint`, that means a user can pass in a "null" uint.
1) This is odd behavior to begin with. Usually a user wants to search by guardianId 12345.
2) The SQL statement (line 96) is going to be weird if the user passes in a null uint.

I think the SQL statement might evaluate as:
```
SELECT * FROM users WHERE guardian_id=0
OR
SELECT * FROM users WHERE guardian_id=NULL;
```
The first statement will never happen. And the second statement will also never return anything because `=NULL` and `IS NULL` are very different things.

So the idea here is to FORCE the user to have to pick a `guardianId`.
userRepo.SelectByGuardianId(uint) so anyone who uses this function must choose a guardianId. They won't give you a "null" guardianId. To do this, just change the signature above to take in uint and then before the SQL statement executes, change the uint into a `NullUint` as a parameter.

ahsu1230: Alternatively, you could keep the signature and change the SQL statement below.
i.e.
if NullUint is valid, do `SELECT * FROM users WHERE guardian_id=?`
Otherwise if invalid, do `SELECT * FROM users WHERE guardian_id IS NULL`.

This is totally valid if we ever want to search for users without a guardian_id... but honestly that's probably not useful at all. So that's why I would opt for the change in signature (use uint as parameter).

ahsu1230:
Looks great! This is a great example of supporting the ability to search by null... but forcing users to only search with a valid content. If we allowed users to search by a null guardianId, it's kind of... not helpful. But at the same time, we should allow null guardianIds because not every student will have a registered guardian.

Great work!

ahsu1230:
lgtm

PR #97, opened 2020-04-04
added publishedAt to programs, locations, sessions

mzhang2021:
@ahsu1230 
@tonywu315


orion/pkg/domains/location.go
```diff
@@ -10,16 +10,17 @@ import (
...
-	CreatedAt time.Time    `json:"-" db:"created_at"`
-	UpdatedAt time.Time    `json:"-" db:"updated_at"`
-	DeletedAt sql.NullTime `json:"-" db:"deleted_at"`
-	LocId     string       `json:"locId" db:"loc_id"`
-	Street    string       `json:"street"`
-	City      string       `json:"city"`
-	State     string       `json:"state"`
-	Zipcode   string       `json:"zipcode"`
-	Room      NullString   `json:"room"`
+	Id          uint
```
ahsu1230, line 13: keep the json stuff from before. we want every field to have a `json:"..."` tag.

orion/pkg/repos/migrations/000010_add_published_at_table_programs.down.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE programs DROP COLUMN published_at;
```
ahsu1230, line 1: could you also add `AFTER deleted_at;`

like in here?
https://github.com/ahsu1230/mathnavigatorSite/pull/104/files

orion/pkg/tests_integration/semester_integration_test.go
```diff
@@ -11,7 +11,7 @@ import (
 
 // Test: Create 3 Semesters and GetAll()
 func Test_CreateSemesters(t *testing.T) {
-	resetTable(t, domains.TABLE_SEMESTERS)
+	resetSessionTables(t)
```
ahsu1230, line 14: Again, kind of weird, that we are reseting a bunch of tables when our test is only operating on the Semesters table.
If things are affecting this test when it shouldn't, I think there may be an underlying bug? Before this test, the SEMESTERS table should be cleared.

And before another test, it would be safe to reset used tables for that test.

ahsu1230:
Everything looks good, except that one part where we have to clear many tables before using basic semester operations. If the clearSessionsTables is required, I feel like there's an underlying bug / unexpected behavior that we're just trying to force cover up.

orion/pkg/router/handler.go
```diff
@@ -41,6 +42,7 @@ func (h *Handler) SetupApiEndpoints() {
 		apiLocations.GET("/v1/location/:locId", controllers.GetLocationById)
 		apiLocations.POST("/v1/location/:locId", controllers.UpdateLocation)
 		apiLocations.DELETE("/v1/location/:locId", controllers.DeleteLocation)
+		apiLocations.POST("/v1/publish", controllers.PublishLocations)
```
tonywu315, line 45: This endpoint should be grouped with those with similar URLS and then sorted by GET, POST, etc. I put my publish endpoints after the GetALl.

orion/pkg/services/location_service.go
```diff
@@ -9,18 +9,20 @@ var LocationService locationServiceInterface = &locationService{}
 
 // Interface for LocationService
 type locationServiceInterface interface {
-	GetAll() ([]domains.Location, error)
+	GetAll(bool) ([]domains.Location, error)
 	GetByLocationId(string) (domains.Location, error)
 	Create(domains.Location) error
 	Update(string, domains.Location) error
 	Delete(string) error
+	GetAllUnpublished() ([]domains.Location, error)
```
tonywu315, line 17: I put this after the GetAll because they are similar.

orion/pkg/tests_integration/publish_integration_test.go
```diff
@@ -37,4 +36,225 @@ func Test_GetUnpublished(t *testing.T) {
 	assert.EqualValues(t, 2021, unpublishedDomains.Achieves[1].Year)
 	assert.EqualValues(t, "message2", unpublishedDomains.Achieves[1].Message)
 	assert.EqualValues(t, 2, len(unpublishedDomains.Achieves))
+
+	resetTable(t, domains.TABLE_ACHIEVEMENTS)
+}
+
+// Test: Get All Unpublished Programs, Publish a Few, then Get All Unpublished Again
+func Test_PublishPrograms(t *testing.T) {
```
tonywu315, line 44: This endpoint is part of programs. `Test_GetUnpublished` should be the only test here.

orion/pkg/controllers/publish_controller_test.go
```diff
@@ -38,3 +65,62 @@ func TestGetUnpublishedAchievements_Success(t *testing.T) {
 	assert.EqualValues(t, "message2", unpublishedDomains.Achieves[1].Message)
 	assert.EqualValues(t, 2, len(unpublishedDomains.Achieves))
 }
+
+func TestGetAllUnpublishedPrograms_Success(t *testing.T) {
+	SetupMock()
+
+	// Create new HTTP request to endpoint
```
tonywu315, line 72: If this test is split into 6, this logic is duplicated 6 times. I think this should either be one test or there should be a helper method that gets the unpublishedDomains once.

orion/pkg/repos/session_repo.go
```diff
@@ -51,7 +53,8 @@ func (sr *sessionRepo) SelectAllByClassId(classId string) ([]domains.Session, er
 			&session.StartsAt,
 			&session.EndsAt,
 			&session.Canceled,
-			&session.Notes); errScan != nil {
+			&session.Notes,
+			&session.PublishedAt); errScan != nil {
```
tonywu315: Since the `published_at` column is after the `deleted_at`, I think this should also be after `&session.DeletedAt` as well as for the other 2 domains.

orion/pkg/repos/session_repo.go
```diff
@@ -121,7 +125,8 @@ func (sr *sessionRepo) Update(id uint, session domains.Session) error {
 		"starts_at=?, " +
 		"ends_at=?, " +
 		"canceled=?, " +
-		"notes=? " +
+		"notes=?, " +
+		"published_at=? " +
```
tonywu315: Same

orion/pkg/controllers/publish_controller_test.go
mzhang2021: That's a good idea.

orion/pkg/tests_integration/publish_integration_test.go
mzhang2021: Not sure what you mean by that. GetUnpublished can only be meaningfully tested if we first publish some programs?

tonywu315: Yes, but each domain has its own endpoint to publish, whereas the GetUnpublished is for all the domains.

mzhang2021: I get that, but testing GetUnpublished independently isn't very meaningful if we don't publish some of the domains and then see if the result changes. Then it would be no different than testing GetAll.

tonywu315: Oh I see. You used both get all unpublished and get all published.

orion/pkg/controllers/location_controller.go
```diff
@@ -88,3 +90,20 @@ func DeleteLocation(c *gin.Context) {
...
+func PublishLocations(c *gin.Context) {
+	// Incoming JSON
+	var locIdsJson []string
+	c.BindJSON(&locIdsJson)
+
+	errors := services.LocationService.Publish(locIdsJson)
+	if len(errors) > 0 {
+		for _, err := range errors {
+			c.Error(err.Error)
+			c.String(http.StatusInternalServerError, err.Error.Error())
```
tonywu315, line 103: I don't think the front-end should get a bunch of errors when they only send one request.

mzhang2021:
DB transactions to be added in separate PR

orion/pkg/controllers/location_controller.go
```diff
@@ -53,6 +55,23 @@ func CreateLocation(c *gin.Context) {
...
```
ahsu1230, line 67: hm it usually doesn't work like this. You can't have multiple error responses to a single request.
A better way to go about it would be to aggregate or concatenate all the errors together.
Then send the aggregate along with one internal server error.

orion/pkg/controllers/program_controller.go
```diff
@@ -53,6 +55,23 @@ func CreateProgram(c *gin.Context) {
...
+func PublishPrograms(c *gin.Context) {
+	// Incoming JSON
+	var programIdsJson []string
+	c.BindJSON(&programIdsJson)
+
+	errors := services.ProgramService.Publish(programIdsJson)
+	if len(errors) > 0 {
+		for _, err := range errors {
+			c.Error(err.Error)
+			c.String(http.StatusInternalServerError, err.Error.Error())
```
ahsu1230, line 67: same as above

orion/pkg/domains/program.go
```diff
@@ -10,15 +9,21 @@ import (
...
+	ProgramId   string    `json:"programId" db:"program_id"`
+	Name        string    `json:"name"`
+	Grade1      uint      `json:"grade1"`
+	Grade2      uint      `json:"grade2"`
+	Description string    `json:"description"`
+}
+
+type ProgramErrorBody struct {
+	ProgramId string `json:"programId"`
+	Error     error  `json:"error"`
```
ahsu1230, line 26: Regarding the above return response for the controllers.
You could use `c.JSON(status, <array of ProgramErrorBody>)`.

orion/pkg/controllers/location_controller.go
ahsu1230: Just like the getAll endpoints, you can use `c.JSON(status, ...)` and create a single object which would be the JSON response body. You can pass in an array of LocationError structs - that would work.

ahsu1230:
almost there!

orion/pkg/controllers/location_controller.go
```diff
@@ -63,8 +64,14 @@ func PublishLocations(c *gin.Context) {
 
 	errorList := services.LocationService.Publish(locIdsJson)
 	if len(errorList) > 0 {
-		c.Error(errors.New("one or more locations failed to publish"))
-		c.JSON(http.StatusInternalServerError, errorList)
+		var errorStrings []string
+		errorStrings = append(errorStrings, "one or more locations failed to publish")
+		for _, errorBody := range errorList {
```
ahsu1230, line 69: Make a public function in `domains` package (you can put it in same file as PublicErrorBody) that concatenates all the errors here.
No need to replace this same code over and over.

So this whole part could look like:
```
err := concatErrors(errorList, "one or more ....")
c.Error(err)
c.String(...)
```

orion/pkg/repos/location_repo.go
```diff
@@ -161,26 +162,28 @@ func (lr *locationRepo) Insert(location domains.Location) error {
...
 		if !location.PublishedAt.Valid {
 			now := time.Now().UTC()
 			location.PublishedAt.Scan(now)
 			err = lr.Update(locId, location)
 			if err != nil {
-				errors = append(errors, domains.LocationErrorBody{LocId: locId, Error: err})
+				errorList = append(errorList, domains.PublishErrorBody{StringId: locId, Error: err})
 			}
+		} else {
+			errorList = append(errorList, domains.PublishErrorBody{StringId: locId, Error: errors.New("location is already published")})
```
tonywu315, line 182: I think we agreed that there should be no error.

orion/pkg/repos/location_repo.go
```diff
@@ -161,26 +162,28 @@ func (lr *locationRepo) Insert(location domains.Location) error {
...
 		location, err := lr.SelectByLocationId(locId)
 		if err != nil {
-			errors = append(errors, domains.LocationErrorBody{LocId: locId, Error: err})
+			errorList = append(errorList, domains.PublishErrorBody{StringId: locId, Error: err})
 			continue
 		}
 		if !location.PublishedAt.Valid {
 			now := time.Now().UTC()
 			location.PublishedAt.Scan(now)
 			err = lr.Update(locId, location)
```
tonywu315, line 177: I think publishing is different from updating. This will also change `updated_at`.

mzhang2021: True. I think I'll replace this whole section with database transactions in a different PR anyways.

orion/pkg/repos/location_repo.go
mzhang2021: I think we agreed that the overall process would still go through, but I don't know if we agreed on whether the individual thing should be treated as an error.
In my opinion it makes sense to treat it as an error, since the option to publish an already published item from the user side shouldn't be available in the first place.

tonywu315: We agreed on idempotency earlier.

mzhang2021: I don't know how such a scenario would arise but ok.

PR #99, opened 2020-04-04
Programs updated

mythicat554:


orion/sites/admin/src/programs/program.js
```diff
@@ -22,24 +22,52 @@ export class ProgramPage extends React.Component {
...
+        </ul>
         <ul id="list-heading">
+          <button className="li-small" onClick={this.onClickSelectAll}>Select All</button>
           <li className="li-med">ProgramKey</li>
           <li className="li-med">Name</li>
           <li className="li-small">Grade1</li>
           <li className="li-small">Grade2</li>
         </ul>
         <ul id="list-rows">
+            <script>hideInputs()</script>
```
ahsu1230, line 56: I don't see a hideInputs() function anywhere?

orion/sites/admin/src/programs/program.styl
```diff
@@ -52,6 +56,22 @@
...
+
 a.add-program
-  display: inline-block
-  margin: spacing2x 0
+    display: inline-block
+    margin: spacing2x 0
+    background-color: colorTurquoise
+    color: white
+    padding: 0.75em;
+    border-radius: 10px;
```
ahsu1230, line 68: keep in mind proper ordering of CSS

display, (padding, margin, border, etc.) then colors

orion/sites/admin/src/programs/program.styl
```diff
@@ -52,6 +56,22 @@
...
+
+a.publish
+    display: inline-block
+    margin: spacing2x 0
+    background-color: colorTurquoise
+    color: white
+    padding: 0.75em;
+    border-radius: 10px;
+    margin-right:10px
```
ahsu1230, line 77: same here

orion/sites/admin/src/programs/program.js
```diff
@@ -22,24 +22,52 @@ export class ProgramPage extends React.Component {
...
     const numRows = rows.length;
-		return (
+    let numUnpublished = 0;
+    let numSelected = 0;
+	return (
       <div id="view-program">
-        <h1>All Programs ({numRows})</h1>
+        <ul>
+            <h1>All Programs ({numRows}) </h1>
+            <div> You have {numUnpublished} unpublished items. </div>
```
ahsu1230, line 45: Maybe you could bundle both of these `<div>s` into a single `<p>`. `<p>` tag is for paragraphs.

orion/sites/admin/src/programs/program.js
mythicat554: oops- this is from when I was trying to hide some checkboxes and I forgot to delete it

orion/sites/admin/src/programs/program.styl
ahsu1230: bump again!

orion/sites/admin/src/programs/program.styl
```diff
@@ -11,6 +11,10 @@
     margin: spacing4x 0
     line-height: lineHeight3
     font-size: fontSize3
+    display: inline-block
```
ahsu1230, line 14: display is usually the highest priority in CSS ordering

orion/sites/admin/src/programs/program.js
```diff
@@ -49,8 +77,16 @@ class ProgramRow extends React.Component {
   render() {
     const row = this.props.row;
     const url = "/program/" + row.programId + "/edit";
+    let isUnpublished = true;
+    let checkbox = <div></div>;
+    if (isUnpublished){
+        checkbox = (<input className="li-small" type="checkbox" name="unpublished" onClick={this.onClickBox}/>);
+    } else {
+        checkbox = (<div className="li-small"></div>);
+    }
```
ahsu1230, line 86: very nice!

ahsu1230: Consider moving this part into a function.
So in render(), you call this:
```
let checkbox = renderCheckbox(true);
```

Somewhere else outside the React Component, the function looks like:
```
function renderCheckbox(isUnpublished) {
    let checkbox = ...
    if (isUnpublished) {...} else {...}
    return checkbox;
}
```

Then in the very near future, instead of just passing true or false, it can be a variable!

orion/sites/admin/src/programs/program.js
```diff
@@ -22,24 +22,52 @@ export class ProgramPage extends React.Component {
...
-        <Link className="add-program" to={"/programs/add"}>Add Program</Link>
+        <ul id="list-buttons">
+            <div className="li-med">
+                <button>
+                <Link className="add-program" to={"/programs/add"}>Add Program</Link>
+                </button>
+            </div>
+            <div className="li-med">
+                <button>
+                <Link className="publish" to={"/programs/add"}>Publish</Link>
```
ahsu1230, line 67: The Publish button won't be a link. It will be an API call like in the edit page, so the button will have an onClick.

For now don't worry about the API call itself, but just make this button use an onClick

orion/sites/admin/src/programs/program.js
```diff
@@ -22,24 +22,52 @@ export class ProgramPage extends React.Component {
...
 
-	render() {
+  onClickSelectAll() {
+      var items = document.getElementsByName("unpublished");
+      for (var i=0; i<items.length; i++) {
+          if (items[i].type == 'checkbox') {
+              items[i].checked = true;
+          }
+      }
+  }
```
ahsu1230, line 32: very nice!

orion/sites/admin/package.json
```diff
@@ -9,7 +9,8 @@
     "build": "webpack --config webpack.prod.js",
     "build-local": "webpack --config webpack.local.js",
     "start": "webpack-dev-server --open --config webpack.local.js",
-    "try": "webpack --display-error-details"
+    "try": "webpack --display-error-details",
+    "prettify": "prettier --config \".prettierrc.yaml\" --write \"**/*.js\""
```
ahsu1230, line 13: Now in the `orion/sites/admin` folder, just call `npm run prettify` and it will do the complex prettify call for you! yayyy

orion/sites/admin/src/programs/program.js
```diff
@@ -21,35 +21,93 @@ export class ProgramPage extends React.Component {
...
         const numRows = rows.length;
+        let numUnpublished = 0;
+        let numSelected = 0;
         return (
             <div id="view-program">
-                <h1>All Programs ({numRows})</h1>
+                <ul>
+                    <h1>All Programs ({numRows}) </h1>
+                    <p>
+                        {" "}
```
ahsu1230, line 49: what are these spaces for?

orion/sites/admin/src/programs/program.js
```diff
@@ -21,35 +21,93 @@ export class ProgramPage extends React.Component {
...
                     <li className="li-med">ProgramKey</li>
                     <li className="li-med">Name</li>
                     <li className="li-small">Grade1</li>
                     <li className="li-small">Grade2</li>
                 </ul>
                 <ul id="list-rows">{rows}</ul>
-                <Link className="add-program" to={"/programs/add"}>
-                    Add Program
-                </Link>
+                <ul id="list-buttons">
```
ahsu1230, line 66: ok in this case, you should make a decision

This is either a "list" in which case you use a `<ul>` or it's a "container" (i.e. buttons-container) in which case it's just a regular `<div>`. If you go with the list, the insides should be `<li>`s. Otherwise divs inside divs are fine. In this case, I would do the divs.

orion/sites/admin/src/programs/program.js
```diff
@@ -21,35 +21,93 @@ export class ProgramPage extends React.Component {
...
+                    <div className="li-med">
+                        <button>
+                            <Link className="add-program" to={"/programs/add"}>
+                                Add Program
+                            </Link>
+                        </button>
+                    </div>
+                    <div className="li-med">
+                        <button
```
ahsu1230, line 75: oh I see what you're trying to do. Ok so get rid of these `<div className="li-med">` stuff.

now give this button the CSS property `position: absolute` while the parent `<div id="buttons-container"/>` gets property `position: relative`.
From here, you can give the button `right: 0` and it will be the "right-most" area of its parent <- who in this case is "buttons-container". This also goes for top, left, and bottom. `top: 0, right: 0` means the upper right most area of the parent.

https://www.w3schools.com/css/css_positioning.asp

orion/sites/admin/src/programs/program.js
mythicat554: I'm not sure, they just appeared after I ran prettify (I'll go ahead and delete them)

orion/sites/admin/src/programs/program.js
```diff
@@ -21,35 +21,89 @@ export class ProgramPage extends React.Component {
...
         const rows = this.state.list.map((row, index) => {
             return <ProgramRow key={index} row={row} />;
         });
         const numRows = rows.length;
+        let numUnpublished = 0;
+        let numSelected = 0;
         return (
             <div id="view-program">
-                <h1>All Programs ({numRows})</h1>
+                <ul>
```
ahsu1230, line 46: There are no lists here, so use `<div>` instead

PR #101, opened 2020-04-05
Introduce React Enzyme component unit testing

ahsu1230:
Unit testing for React. Enzyme is created by AirBnb and is a React library dedicated to testing. Jest is a Javascript Testing framework. Together, they work together to help us render our React components and test that things do indeed work as intended.

For now, we'll be using Enzyme's React-shallow rendering which only calls the `render()` function with props or state values. The Deep rendering will actually call `componentDidMount` which we have API stuff in, but we'll get into that later.
https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/some.html

To use Jest, refer to this:
https://jestjs.io/docs/en/expect

But basically, it would be great if we had unit tests for every component in our codebase. This is simply to make sure that whenever we change a component it doesn't automatically break another component. It also helps us catch silly syntax errors - especially in Javascript that is weakly typed and won't tell us until we manually try things ourselves.

When testing, for now, it's just good to get the basics. As in tests for basic rendering, testing props and states, and parent-children relationships (again, this is all available in Enzyme shallow rendering).

To run unit tests, simply run:
```
npm run test
```
If you want to be fancy, you can run `npm run test:watch` and as you change your javascript files, it will automatically re-run tests for the files that are affected. So you can see exactly what lines you make are breaking.

@mythicat554 @MiniViolin if you have read this, please approve this PR :)


orion/sites/admin/src/achieve/achieve.test.js
```diff
@@ -0,0 +1,41 @@
...
+
+describe("test", () => {
+    const component = shallow(<AchievePage />);
+
+    test("renders", () => {
+        expect(component.exists()).toBe(true);
+        expect(component.find("h1").text()).toContain("All Achievements");
+        expect(component.find("Link").text()).toContain("Add Achievement");
+        expect(component.find("AchieveRow").length).toBe(0);
+    });
```
ahsu1230, line 13: The above is the "main" test we really want to make sure we have for every Component. This ensures that the component will always render. If there's a typo or a syntax anywhere, the component will not render and thus this test will fail.

orion/sites/admin/src/achieve/achieve.test.js
```diff
@@ -0,0 +1,41 @@
...
+        let row0 = component.find("AchieveRow").at(0);
+        expect(row0.prop("achieve")).toHaveProperty("Id", 1);
+        expect(row0.prop("achieve")).toHaveProperty("year", 2020);
+        expect(row0.prop("achieve")).toHaveProperty("message", "Awesome");
+
+        let row1 = component.find("AchieveRow").at(1);
+        expect(row1.prop("achieve")).toHaveProperty("Id", 2);
+        expect(row1.prop("achieve")).toHaveProperty("year", 2019);
+        expect(row1.prop("achieve")).toHaveProperty("message", "Possum");
+    });
```
ahsu1230, line 40: The above test is how you would test a parent Component sending data through `this.props.xxxx` to children Components.

orion/sites/admin/src/achieve/achieveEdit.test.js
```diff
@@ -0,0 +1,79 @@
...
+    test("renders", () => {
+        expect(component.exists()).toBe(true);
+        expect(component.find("h2").text()).toContain("Add Achievement");
+        expect(component.find("h4").at(0).text()).toBe("Year");
+        expect(component.find("h4").at(1).text()).toBe("Message");
+        expect(component.find("h4").length).toBe(2);
+        expect(component.find("button").at(0).text()).toBe("Save");
+        expect(component.find("button").at(1).text()).toBe("Cancel");
+        expect(component.find("button").length).toBe(2);
+    });
```
ahsu1230, line 18: Again make sure every Component gets a basic render test

PR #102, opened 2020-04-05
Add babel class-properties plugin

ahsu1230:
With this plugin, we can get away from using the React constructor. We no longer need to declare a constructor with the states (can simply do the arrow thing) and we don't need to bind functions in the constructor anymore either. Yay!

Just an improvement to QualityOfLife. I also took the liberty of refactoring out modal and deleteButton rendering into a function.
Since this function is probably used many times throughout our pages... maybe we could considering create a component Util function that gets used by multiple components.

If you've read this @MiniViolin @mythicat554 please approve this PR :). It would be great if we eventually moved to only using this style so we keep a consistent styling.


orion/sites/admin/src/location/location.js
```diff
@@ -1,17 +1,13 @@
 "use strict";
 require("./location.styl");
 import React from "react";
-import ReactDOM from "react-dom";
 import API from "../api.js";
 import { Link } from "react-router-dom";
 
 export class LocationPage extends React.Component {
-    constructor(props) {
```
ahsu1230, line 9: no more constructor

orion/sites/admin/src/location/locationEdit.js
```diff
@@ -1,36 +1,22 @@
...
-        };
-        this.handleChange = this.handleChange.bind(this);
-
-        this.onClickCancel = this.onClickCancel.bind(this);
-        this.onClickDelete = this.onClickDelete.bind(this);
-        this.onClickSave = this.onClickSave.bind(this);
-
-        this.onDeleted = this.onDeleted.bind(this);
-        this.onSaved = this.onSaved.bind(this);
-        this.onDismissModal = this.onDismissModal.bind(this);
```
ahsu1230, line 32: no more binds yes!

orion/sites/admin/src/location/locationEdit.js
```diff
@@ -88,71 +74,41 @@ export class LocationEditPage extends React.Component {
...
+    onDeleted = () => {
         const locId = this.props.locId;
         API.delete("api/locations/v1/location/" + locId).then((res) => {
             window.location.hash = "locations";
         });
-    }
+    };
 
-    onSaved() {
+    onSaved = () => {
```
ahsu1230, line 86: do this arrow thingy! It basically means a function that takes in no parameters does this ` => {...}`

orion/sites/admin/src/location/locationEdit.js
```diff
@@ -44,26 +30,26 @@ export class LocationEditPage extends React.Component {
...
-                    inputRoom: location.room,
+                    inputRoom: location.room || "",
                     isEdit: true,
                 });
             });
         }
     }
 
-    handleChange(event, value) {
+    handleChange = (event, value) => {
```
ahsu1230, line 40: Here's an example of arrow functions with parametres

PR #107, opened 2020-04-10
Publish classes

tonywu315:
@ahsu1230 
@mzhang2021


orion/pkg/tests_integration/announce_integration_test.go
```diff
@@ -11,7 +11,7 @@ import (
 
 // Test: Create 3 Announcements and GetAll()
 func Test_CreateAnnouncements(t *testing.T) {
-	resetTable(t, domains.TABLE_ANNOUNCEMENTS)
+	resetAllTables(t)
```
ahsu1230, line 14: I find it odd we have to reset ALL tables for announcement tests... are you sure we have to do that?
It would be great to reset only announcements table in these tests since these tests shouldn't affect any other integration test.
If it does for some reason, it looks like we have an underlying bug somewhere!

I think we can keep resetTable() but you can move it to setupTests.go

orion/pkg/controllers/base_controller_test.go
```diff
@@ -45,19 +45,24 @@ var classService mockClassService
...
+	mockGetAll                    func(bool) ([]domains.Class, error)
+	mockGetUnpublished            func() ([]domains.Class, error)
 	mockGetByClassId              func(string) (domains.Class, error)
 	mockGetByProgramId            func(string) ([]domains.Class, error)
 	mockGetBySemesterId           func(string) ([]domains.Class, error)
 	mockGetByProgramAndSemesterId func(string, string) ([]domains.Class, error)
 	mockCreate                    func(domains.Class) error
 	mockUpdate                    func(string, domains.Class) error
 	mockDelete                    func(string) error
+	mockPublish                   func([]string) error
```
mzhang2021, line 57: Return list of errors

mzhang2021, line 57: Return list of errors

tonywu315: I think this pull request is big enough already. I'll put this in another pull request.

PR #108, opened 2020-04-10
Added search, pageSize, and offset

tonywu315:
@ahsu1230 
@mzhang2021

### DIFF
--- a/orion/pkg/repos/user_repo.go
+++ b/orion/pkg/repos/user_repo.go
@@ -38,7 +38,7 @@ func (ur *userRepo) SelectAll(search string, pageSize, offset int) ([]domains.Us
 	if getAll {
 		query = "SELECT * FROM users LIMIT ? OFFSET ?"
 	} else {
-		query = "SELECT * FROM users WHERE ? IN (first_name, last_name, middle_name) LIMIT ? OFFSET ?"
+		query = "SELECT * FROM users WHERE ? IN (first_name,last_name,middle_name) LIMIT ? OFFSET ?"
 	}
 	stmt, err := ur.db.Prepare(query)
 	if err != nil {
--- a/orion/pkg/repos/user_repo.go
+++ b/orion/pkg/repos/user_repo.go
@@ -33,8 +33,9 @@ func (ur *userRepo) Initialize(db *sql.DB) {
 func (ur *userRepo) SelectAll(search string, pageSize, offset int) ([]domains.User, error) {
 	results := make([]domains.User, 0)
 
+	getAll := len(search) == 0
 	var query string
-	if len(search) == 0 {
+	if getAll {
 		query = "SELECT * FROM users LIMIT ? OFFSET ?"
 	} else {
 		query = "SELECT * FROM users WHERE ? IN (first_name, last_name, middle_name) LIMIT ? OFFSET ?"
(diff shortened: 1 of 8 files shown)


orion/pkg/repos/user_repo.go
```diff
@@ -29,15 +29,27 @@ func (ur *userRepo) Initialize(db *sql.DB) {
...
+	}
+	stmt, err := ur.db.Prepare(query)
 	if err != nil {
 		return nil, err
 	}
 	defer stmt.Close()
-	rows, err := stmt.Query()
+
+	var rows *sql.Rows
+	if len(search) == 0 {
```
ahsu1230, line 48: nit: you duplicate logic here and line 36 (`len(search == 0)`)
You could define a boolean value above and use it both here and line the if/else block in 36.

orion/pkg/repos/user_repo.go
```diff
@@ -29,15 +29,27 @@ func (ur *userRepo) Initialize(db *sql.DB) {
...
-func (ur *userRepo) SelectAll() ([]domains.User, error) {
+func (ur *userRepo) SelectAll(search string, pageSize, offset int) ([]domains.User, error) {
 	results := make([]domains.User, 0)
 
-	stmt, err := ur.db.Prepare("SELECT * FROM users")
+	var query string
+	if len(search) == 0 {
+		query = "SELECT * FROM users LIMIT ? OFFSET ?"
+	} else {
+		query = "SELECT * FROM users WHERE ? IN (first_name, last_name, middle_name) LIMIT ? OFFSET ?"
```
ahsu1230, line 39: very nice!

orion/pkg/repos/user_repo_test.go
```diff
@@ -28,7 +28,7 @@ func TestSelectAllUsers(t *testing.T) {
 	// Mock DB statements and execute
 	rows := getUserRows()
 	mock.ExpectPrepare("^SELECT (.+) FROM users").ExpectQuery().WillReturnRows(rows)
-	got, err := repo.SelectAll()
+	got, err := repo.SelectAll("", 100, 0)
```
ahsu1230, line 31: can we have an additional user_repo_test that tests:
```
mock.ExpectPrepare("^SELECT (.+) FROM users WHERE ___ IN (...)")
```
This should get called when we call repo.SelectAll("something", ...).

Thanks!

orion/pkg/tests_integration/user_integration_test.go
```diff
@@ -30,6 +30,37 @@ func Test_CreateUsers(t *testing.T) {
 	assert.EqualValues(t, 3, len(users))
 }
 
+// Test: Create 3 Users and GetAll()
```
ahsu1230, line 33: Create 3 Users and `search by pagination`

orion/pkg/tests_integration/user_integration_test.go
```diff
@@ -30,6 +30,37 @@ func Test_CreateUsers(t *testing.T) {
...
+
+	// Validate results
+	assert.EqualValues(t, http.StatusOK, recorder2.Code)
+	var users2 []domains.User
+	if err := json.Unmarshal(recorder2.Body.Bytes(), &users2); err != nil {
+		t.Errorf("unexpected error: %v\n", err)
+	}
+	assertUser(t, 3, users2[0])
+	assert.EqualValues(t, 1, len(users2))
+}
```
ahsu1230, line 62: very nice!

PR #109, opened 2020-04-10
Added parallelism

tonywu315:
@ahsu1230 
@mzhang2021

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,7 @@ jobs:
           MYSQL_DATABASE: mathnavdb_test
           MYSQL_USER: ci_tester
           MYSQL_PASSWORD: test
+    parallelism: 4
 
     working_directory: /go/src/github.com/ahsu1230/mathnavigatorSite
     steps:


ahsu1230:
Well it is definitely correct but doesn't seem like the tests are going any faster. 
This was more of a research task. And honestly, maybe we don't have enough tests to make parallel testing worth it yet. If it doesn't help, then we can close this task for now until the benefits of it are more apparent. Parallel testing is great but it risks "flaky" tests, which means there are rare occurrences that parallelism can introduce random bugs and only some of the times...

This is more of a research task. Maybe play around with this:
https://circleci.com/docs/2.0/parallelism-faster-jobs/#using-the-circleci-cli-to-split-tests
One simple example could be do split between regular orion tests and the integration tests since those are supposed to be very separate isolated groups.

ahsu1230:
Let's close this PR for now. I don't think it benefits us atm. Maybe one day when we have a billion more tests, but thank you for looking into it!

ahsu1230:
And we can always reference to it later.

PR #110, opened 2020-04-11
Locations updated

mythicat554:
I didn't finish figuring out how to update the number of checkboxes checked off yet, but I created the checkboxes, the check all, the publish button, etc.


ahsu1230:
Click on the "Details" link next to the CircleCi build check to see how your build failed. :)

orion/sites/admin/src/location/location.styl
```diff
@@ -6,12 +6,23 @@
     padding: 0 spacing2x
     box-sizing: border-box
     color: colorBlackDark
+    position: relative
```
ahsu1230, line 9: position: relative goes right below `display` for CSS ordering

orion/sites/admin/src/location/location.js
```diff
@@ -20,29 +20,79 @@ export class LocationPage extends React.Component {
         });
     }
 
+    onClickSelectAll() {
+        var items = document.getElementsByName("unpublished");
+        for (var i = 0; i < items.length; i++) {
+            if (items[i].type == "checkbox") {
+                items[i].checked = true;
+            }
+        }
```
ahsu1230, line 29: When a user selects all, maybe here is the time to modify this.state.numSelected;

orion/sites/admin/src/location/location.js
```diff
@@ -20,29 +20,79 @@ export class LocationPage extends React.Component {
...
+        console.log("clicked publish");
+    }
+
     render() {
         const location = this.state.list.map((location, index) => {
             return <LocationRow key={index} location={location} />;
         });
         const numLocations = location.length;
+        let numUnpublished = 0;
+        let numSelected = 0;
```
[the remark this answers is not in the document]
mythicat554: I'm a bit stumped because the checkbox is located in a different class, and I'm not sure how I would change my variable in the LocationPage from a button being clicked in the LocationRow class. I tried making an instance of the classes to try and call functions between the two, but it's not working out :(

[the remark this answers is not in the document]
ahsu1230: Oh I see. Ok so if I'm understanding your problem at the moment, the checkbox is in `LocationRow` while the data (numSelected) is in `LocationPage`.
Remember how we did modals? We passed in a function to a Modal to be used when something happens (this is called callbacks if you remember). You can do something similar here. Define a function `onCheckRow`. It's defined in LocationPage. When you create LocationRow, pass in that function so it will look like this:

```
<LocationRow key={index} location={location} onCheckRow={this.onCheckRow}/>;
```

Once you do that, this onCheckRow function is available inside the LocationRow component! You can access it by using `this.props.onCheckRow`. I would probably an onClick of the checkbox to this new function. So it would look something like this:
```
<input type="checkbox" onClick={this.props.onCheckRow}>
```

So what I would do first is set this up so that `onCheckRow` is super simple like just console.log something.
When I click on a checkbox, we should see that print statement in the console.

orion/sites/admin/src/location/location.js
```diff
@@ -20,29 +24,97 @@ export class LocationPage extends React.Component {
...
 class LocationRow extends React.Component {
+    renderCheckbox(isUnpublished) {
+        let checkbox = <div> </div>;
+        if (isUnpublished) {
+            return (checkbox = (
+                <input
+                    className="li-small"
+                    type="checkbox"
+                    name="unpublished"
+                    id="unpublished"
```
ahsu1230, line 109: I think this might be the problem. id's are used for main sections and are SUPPOSED to only be for elements that appear once per page. This is a checkbox, so you'll have many of them - you want to use a className.

After that, when you search for all checkboxes, I think you can search by Class instead of search by Id.

orion/sites/admin/src/location/location.js
```diff
@@ -20,29 +24,97 @@ export class LocationPage extends React.Component {
         });
     }
 
+    onCheckRow() {
+        if (document.getElementById("unpublished").checked) {
```
ahsu1230, line 28: I thiinnk the problem could be here where you're searching by Id instead of searching by class.

Turns out document.getElementByX can return an array of elements so you'll need to loop over them.
`document.getElementsByClassName('test')[0]`

However, if `document.getElementsByClassName(...)` doesn't find anything, it will return `undefined`.
So I would do something like this:
```
const elements = document.getElementsByClassName(...) || [];
... then loop
```
What this does is if `document.blahblah` is a valid array, use that value! If it becomes undefined, assign it to `[]` instead. This overall idea is a Javascript thing called `falsy values`.

orion/sites/admin/src/location/location.js
```diff
@@ -20,29 +24,97 @@ export class LocationPage extends React.Component {
...
+            this.setState({ numSelected: this.state.numSelected + 1 });
+        } else {
+            this.setState({ numSelected: this.state.numSelected - 1 });
```
ahsu1230, line 31: And oh yea, if you're going to loop, make sure to call this.setState() once AFTER the loop finishes.
this.setState() is actually an asynchronous call meaning it will take a while (several milliseconds) until React actually updates the state variables to then rerender data.

So rather than have 5 asynchronous calls to increment some variable by 1 at a time.
It's better to call one asynchronous call, and simply increment by 5.
Hope that makes sense.

mythicat554:
@ahsu1230  I think I got the checkboxes to work?

orion/sites/admin/src/location/location.js
```diff
@@ -20,29 +24,100 @@ export class LocationPage extends React.Component {
...
-                </button>
+                <div id="list-buttons">
+                    <button>
+                        <Link to={"/locations/add"} id="add-location">
+                            Add Location
+                        </Link>
+                    </button>
+                    <button
+                        id="publish"
+                        className="publish"
```
ahsu1230, line 93: Does this button need to have both `id` and `className`? Probably just className right?

orion/sites/admin/src/location/location.js
```diff
@@ -20,29 +24,100 @@ export class LocationPage extends React.Component {
...
 class LocationRow extends React.Component {
+    renderCheckbox(isUnpublished) {
+        let checkbox = <div> </div>;
+        if (isUnpublished) {
+            return (checkbox = (
+                <input
+                    className="li-small"
+                    type="checkbox"
+                    name="unpublished"
+                    id="unpublished"
```
ahsu1230, line 112: same here, you can remove id

ahsu1230:
lgtm!

PR #113, opened 2020-04-16
Database transactions for programs, locations, sessions

mzhang2021:
@ahsu1230 
@tonywu315

### DIFF
--- a/orion/pkg/repos/program_repo.go
+++ b/orion/pkg/repos/program_repo.go
@@ -2,6 +2,7 @@ package repos
 
 import (
 	"database/sql"
+	"errors"
 	"github.com/ahsu1230/mathnavigatorSite/orion/pkg/domains"
 	"time"
 )
@@ -21,7 +22,7 @@ type ProgramRepoInterface interface {
 	SelectAllUnpublished() ([]domains.Program, error)
 	SelectByProgramId(string) (domains.Program, error)
 	Insert(domains.Program) error
-	Publish([]string) ([]domains.PublishErrorBody, error)
+	Publish([]string) error
 	Update(string, domains.Program) error
 	Delete(string) error
 }
@@ -163,16 +164,16 @@ func (pr *programRepo) Insert(program domains.Program) error {
 	return handleSqlExecResult(execResult, 1, "program was not inserted")
 }
 
-func (pr *programRepo) Publish(programIds []string) ([]domains.PublishErrorBody, error) {
-	errorList := make([]domains.PublishErrorBody, 0)
+func (pr *programRepo) Publish(programIds []string) error {
+	var errorString string
 
 	tx, err := pr.db.Begin()
 	if err != nil {
-		return nil, err
+		return err
 	}
 	stmt, err := tx.Prepare("UPDATE programs SET published_at=? WHERE program_id=? AND published_at IS NULL")
 	if err != nil {
-		return nil, err
+		return err
 	}
 	defer stmt.Close()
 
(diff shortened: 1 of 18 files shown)


orion/pkg/repos/program_repo.go
```diff
@@ -163,26 +163,34 @@ func (pr *programRepo) Insert(program domains.Program) error {
...
 
+	tx, err := pr.db.Begin()
+	if err != nil {
+		return nil, err
+	}
+	stmt, err := tx.Prepare("UPDATE programs SET published_at=? WHERE program_id=? AND published_at IS NULL")
+	if err != nil {
+		return nil, err
+	}
+	defer stmt.Close()
```
mzhang2021, line 177: In addition to specific errors per program id, the function can also return general errors that don't pertain to a specific id (i.e. tx.Prepare statement fails). I'm not sure if returning two items is the best way to deal with this.
The alternative is to concatenate the errors immediately on the repo level instead of waiting until later, but then the PublishErrorBody struct becomes obsolete. I think that might be better though. What do yall think?

tonywu315: I did the latter. The PublishErrorBody isn't obsolete because it removes the need to create a separate struct for each one.

mzhang2021: I'm saying we don't need any struct at all if we do it the latter way.

orion/pkg/repos/repos_util.go
```diff
@@ -27,3 +28,16 @@ func handleSqlExecResult(result sql.Result, expected int64, errorMessage string)
 	}
 	return nil
 }
+
+func appendError(errorString string, stringId string, rowId uint, err error) string {
```
ahsu1230, line 32: I think this function would be easier to use if the signature just had a generic string `id`, so 3 params instead of 4. In the case we need to pass a uint, just turn the id into a string. We're simply appending error strings so it's fine to cast to string in this case.

```
appendError(errorString, fmt.Sprint(rowId), err)
```

orion/pkg/repos/program_repo_test.go
```diff
@@ -222,6 +222,35 @@ func TestInsertProgram(t *testing.T) {
...
+	// Mock DB statements and execute
+	result := sqlmock.NewResult(1, 1)
+	mock.ExpectBegin()
+	mock.ExpectPrepare("^UPDATE programs SET published_at=(.*) WHERE program_id=(.*)  AND published_at IS NULL").
+		ExpectExec().
+		WithArgs(sqlmock.AnyArg(), "prog1").
+		WillReturnResult(result)
+	mock.ExpectExec("^UPDATE programs SET published_at=(.*) WHERE program_id=(.*)  AND published_at IS NULL").
+		WithArgs(sqlmock.AnyArg(), "prog2").
+		WillReturnResult(result)
```
ahsu1230, line 241: great test!

orion/pkg/repos/repos_util.go
```diff
@@ -27,3 +28,16 @@ func handleSqlExecResult(result sql.Result, expected int64, errorMessage string)
...
+	if len(stringId) > 0 {
+		errorString += stringId + ": "
+	}
+	if rowId > 0 {
+		errorString += fmt.Sprint(rowId) + ": "
+	}
+	if err != nil {
+		errorString += err.Error() + "\n"
```
ahsu1230, line 40: It would be great if in the error message, you made it more complete. Remember, you're using error messages to help you debug erros in the future, so it's good to be clear what's happening.

I would do:
```
"id: " + id + " error: " + err.Error() + ", "
```
Or something like that. So when you actually see the error message you know what values are being mapped to what keys.

PR #114, opened 2020-04-17
implemented bulk create and delete

mzhang2021:
@ahsu1230 
@tonywu315


orion/pkg/controllers/session_controller.go
```diff
@@ -35,21 +37,29 @@ func GetSessionById(c *gin.Context) {
...
-		c.String(http.StatusBadRequest, err.Error())
-		return
+	var errorString string
+	for _, session := range sessionsJson {
+		if err := session.Validate(); err != nil {
+			errorString = appendError(errorString, fmt.Sprint(session.Id), err)
+		} else {
+			validatedSessions = append(validatedSessions, session)
+		}
 	}
```
mzhang2021, line 52: Actually now that I think about it, would it be cleaner to move validation to the repo layer as well instead of iterating over the list of sessions twice?

ahsu1230: I think that's fair. Make sure to call the validation function before trying to persist to db (Calling the Exec statement).

orion/pkg/tests_integration/session_integration_test.go
```diff
@@ -149,7 +143,8 @@ func Test_DeleteSession(t *testing.T) {
 	assert.EqualValues(t, http.StatusOK, recorder5.Code)
 
 	// Delete
-	recorder6 := sendHttpRequest(t, http.MethodDelete, "/api/sessions/v1/session/1", nil)
+	body6 := createJsonBody([]uint{1})
+	recorder6 := sendHttpRequest(t, http.MethodDelete, "/api/sessions/v1/delete", body6)
```
ahsu1230, line 147: can we test 2 sessions getting deleted in tihs test? (array of size 2)

PR #116, opened 2020-04-18
Dashboard api

mythicat554:


orion/sites/admin/src/home/home.js
```diff
@@ -2,30 +2,210 @@
...
+    }
+
+    onClickPrograms() {
+        this.setState({ noUnpub: "" });
+        let programs = this.state.programs.map((row, index) => {
+            return <DashboardRow key={index} row={row} />;
+        });
+        this.setState({ row: programs });
+        this.setState({ selection: "programs" });
+        this.setState({ num: programs.length });
```
ahsu1230, line 52: it looks like num is always equal to `this.state.row.length` right? 
In this case, just use that instead of having to keep track of another state variable (num). You can easily get into a state where you forget to update num and then your states become messed up.

orion/sites/admin/src/home/home.js
```diff
@@ -2,30 +2,210 @@
...
+        this.state.switch = true;
+        if (programs.length == 0) {
+            this.setState({
+                noUnpub: "All items are published! Your website is up to date!",
+            });
+            this.state.switch = false;
```
ahsu1230, line 58: remember to modify a state variable, put it into this.setState. To update multiple state variables, you can do:
```
this.setState({
    noUnpub: ...,
    switch: false
});
```

orion/sites/admin/src/home/home.js
```diff
@@ -2,30 +2,210 @@
...
+        this.setState({ row: classes });
+        this.setState({ selection: "classes" });
+        this.setState({ num: classes.length });
+        this.state.switch = true;
+        if (classes.length == 0) {
+            this.setState({
+                noUnpub: "All items are published! Your website is up to date!",
+            });
+            this.state.switch = false;
+        }
```
ahsu1230, line 76: I would do ONE `this.setState()` call at the end of this function.
You can update multiple state values by providing a single object.
```
this.setState({
    noUnpub: ...,
    switch: false,
    rows: ...,
    etc.
});
```

orion/sites/admin/src/home/home.js
```diff
@@ -2,30 +2,210 @@
...
+    constructor(props) {
+        super(props);
+        this.state = {
+            programs: [],
+            classes: [],
+            locations: [],
+            achieves: [],
+            semesters: [],
+            sessions: [],
+            row: [],
```
ahsu1230, line 17: nit: rename this to `rows` or `dashboardRows`

orion/sites/admin/src/home/home.js
```diff
@@ -2,30 +2,210 @@
...
             </div>
         );
     }
 }
+
+class DashboardRow extends React.Component {
+    render() {
+        const row = this.props.row;
+        return (
+            <div id="dashboard-row">
```
ahsu1230, line 199: You're going to have many `dashboard-row`s right? So make this a className not id

orion/sites/admin/src/home/home.js
```diff
@@ -2,30 +2,210 @@
...
+                <div>
+                    {row.name}
+                    {row.classId}
+                    {row.locId}
+                    {row.message}
+                    {row.semesterId}
+                    {row.sessionId}
+                </div>
```
ahsu1230, line 207: extra `<div>` that's unneeded?

orion/sites/admin/src/home/home.styl
```diff
@@ -30,10 +27,16 @@
...
+        padding: spacing1x
         border: 2px solid black
         margin-top: spacing2x
         margin-bottom: 60px
 
+#view-home #dashboard-row div
+    border-bottom: 1px solid colorGrayDark
+    margin: spacing1x
+
+
```
ahsu1230, line 39: nit: extra space

orion/sites/admin/src/home/home.js
```diff
@@ -2,30 +2,210 @@
...
                     <button id="go-to-page">Go to Page</button>
                 </div>
             </div>
         );
     }
 }
+
+class DashboardRow extends React.Component {
+    render() {
+        const row = this.props.row;
```
ahsu1230, line 197: Consider another way of doing this. Instead of passing an entier `row` to `DashboardRow`, pass in just a single `title` string variable and display it below. In this scenario, you only have to worry about displaying one variable.

Above, in your `DashboardRow` creating map part, you can specify which variable of the object you want to pass.
```
let semesters = this.state.semesters.map((row, index) => {
            return <DashboardRow key={index} title={row.semesterId} />;
});
```

I think the concern with the current approach is you end up risking displaying multiple things when you don't intend to.

orion/sites/admin/src/home/home.js
```diff
@@ -2,30 +2,210 @@
...
                     </ul>
                 </div>
                 <div id="box-and-button">
-                    <div className="boxed">text</div>
+                    <div className="boxed">
+                        {unpubMessage}
+                        {this.state.row}
+                        {this.state.noUnpub}
+                    </div>
                     <button id="go-to-page">Go to Page</button>
```
ahsu1230, line 188: I think this GoToPage button needs an onClick?

mythicat554:
@ahsu1230 I believe I've fixed everything you commented (hopefully)

orion/sites/admin/src/home/home.js
```diff
@@ -2,30 +2,211 @@
...
+                    </button>
                 </div>
             </div>
         );
     }
 }
+
+class DashboardRow extends React.Component {
+    render() {
+        const rows = this.props.title;
```
ahsu1230, line 209: nit: rename this to title

orion/sites/admin/src/home/home.js
```diff
@@ -2,30 +2,211 @@
...
+        if (sessions.length == 0) {
+            this.setState({
+                noUnpub: "All items are published! Your website is up to date!",
+                switch: false,
+            });
+        }
+    }
+
+    onClickPage() {
+        console.log("Go to Page clicked");
```
ahsu1230, line 153: do you think you can actually link to the next page. We've done it before in the other pages.
You use something like window.location.___ = "programs";

Note you do know where to link to based on the this.state.selection variable.

orion/sites/admin/src/home/home.js
```diff
@@ -2,30 +2,211 @@
...
+            rows: sessions,
+            selection: "sessions",
+            switch: true,
+        });
+        if (sessions.length == 0) {
+            this.setState({
+                noUnpub: "All items are published! Your website is up to date!",
+                switch: false,
+            });
+        }
```
ahsu1230, line 149: This looks so much neater. Here's one more thing you could do:
Ideally, it would be great to have only ONE this.setState(...). And you can do so by using this fancy one-line if else notation.

It works like this:
`<CONDITION> ? <result if true> : <result if false>`
This is also a thing in Java too actually and a few other languages. It's just an easy way to do a one-line if/else statement and you can assign the result to a variable.

So your code can be like this:
```
let noUnpub = sessions.length == 0 ? "..." : "";
let switch = sessions.length == 0 ? false : true
```

And now you can use one this.setState(...) after these two statements.

mythicat554: wow this fancy if else notation is pretty nice! I put them inside the setState still though.

orion/sites/admin/src/home/home.js
```diff
@@ -2,30 +2,195 @@
...
+            return <DashboardRow key={index} title={row.name} />;
+        });
+        this.setState({
+            noUnpub:
+                programs.length == 0
+                    ? "All items are published! Your website is up to date!"
+                    : "",
+            rows: programs,
+            selection: "programs",
+            switch: programs.length == 0 ? false : true,
```
ahsu1230, line 56: I guess technically this can just be `switch: programs.length != 0` but at least you got try out the `? :` operator!

PR #121, opened 2020-04-20
Crypto-hash Utility Class

mzhang2021:
@ahsu1230 
@tonywu315

### DIFF
--- a/orion/pkg/domains/hash.go
+++ b/orion/pkg/domains/hash.go
@@ -1,0 +1,23 @@
+package domains
+
+import (
+	"golang.org/x/crypto/bcrypt"
+)
+
+type Hash struct {
+	HashBytes []byte
+}
+
+func (hash *Hash) GetHash(str string) error {
+	bytes, err := bcrypt.GenerateFromPassword([]byte(str), bcrypt.DefaultCost)
+	if err != nil {
+		return err
+	}
+	hash.HashBytes = bytes
+	return nil
+}
+
+func (hash *Hash) Compare(str string) (bool, error) {
+	err := bcrypt.CompareHashAndPassword(hash.HashBytes, []byte(str))
+	return err == nil, err
+}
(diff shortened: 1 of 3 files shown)


orion/pkg/domains/hash_test.go
```diff
@@ -0,0 +1,33 @@
+package domains
+
+import (
+	"testing"
+)
+
+func TestHash(t *testing.T) {
+	var hash Hash
+	err := hash.GetHash("potato")
```
ahsu1230, line 9: hmm I think the usage would make more sense if we had something like this:
`hash, err := CreateHash("potato")`

It would be great if I had a utility function that actually created a hashed password from a string.

orion/pkg/domains/hash_test.go
```diff
@@ -0,0 +1,33 @@
...
+	if err != nil {
+		t.Errorf("Error: %v", err)
+	}
+
+	if match, _ := hash.Compare("pineapple"); match {
```
ahsu1230, line 14: I think this usage is fine.

orion/pkg/domains/hash.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"golang.org/x/crypto/bcrypt"
+)
+
+type Hash struct {
+	HashBytes []byte
+}
+
+func (hash *Hash) GetHash(str string) error {
```
ahsu1230, line 11: I think having Compare be part of the Hash struct makes sense. But for creating, I don't think it should be a struct method. It can be a standalone utility function that returns a hash, and/or an error. See my comment below in the test file.

orion/pkg/domains/hash_test.go
mzhang2021: > It would be great if I had a utility function that actually created a hashed password from a string.
So like generate a password for the user to use as their actual password?

PR #122, opened 2020-04-20
Added Documentation

tonywu315:
@ahsu1230 
@mzhang2021

### DIFF
--- a/resources/backend/json.md
+++ b/resources/backend/json.md
@@ -20,13 +20,28 @@ JSON deserializing is the opposite. When frontend sends data to the backend in t
 
 which takes in a slice of bytes and inserts them into the `v` parameter.
 
+# Golang Tags
+
+In Golang, field declarations in a struct may be followed by an optional [string literal tag](https://golang.org/ref/spec#Struct_types). When serializing structs into JSON, we can specify the name of the field in JSON. 
+
+Here is an example:
+```go
+type Example struct {
+	Id        uint      `json:"id"`
+	Empty     string    `json:"empty,omitempty"`
+	CreatedAt time.Time `json:"-"`
+}
+```
+
+In this struct, when we serialize, the `Id` field will have a JSON field of `id`. The `omitempty` means that if `Empty` is empty, it will not show up in the JSON. `-` means that this field will be omitted from the JSON no matter what.
+
 # Uses in Our Codebase
 
 In our codebase, [gin](https://github.com/ahsu1230/mathnavigatorSite/blob/master/resources/backend/04_gin.md) handles the JSON serializing and deserialization for us. However, for testing, we need to use `Marshal` and `Unmarshal`.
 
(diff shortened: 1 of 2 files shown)


resources/backend/sql_advanced.md
```diff
@@ -5,11 +5,78 @@
...
+CREATE TABLE foreign_example
+(
+    id         int,
+    example_id int,
+    PRIMARY KEY (id),
+    FOREIGN KEY (example_id) REFERENCES example (id)
+);
+```
+
+A row can be inserted into `foreign_example` normally, but for the `example_id` column, the corresponding `id` should be used. Once a row is inserted, the `example` row it references cannot be updated or deleted. Attempting to do so yields
```
ahsu1230, line 70: Add one more sentence about what would the correct way be to delete a row referencing a foreign key?

If you want to delete example.id 22, you must first delete all rows that reference to it (aka delete all rows in foreign_example that reference example.id 22). Once all of those are deleted, then you can delete example.id 22.

ahsu1230: Actually,  explain that concept after describing the below error messages

ahsu1230: You could also give examples of how we use foreign keys in our codebase (like what you did in the other file)

resources/backend/json.md
```diff
@@ -0,0 +1,34 @@
...
+In our codebase, [gin](https://github.com/ahsu1230/mathnavigatorSite/blob/master/resources/backend/04_gin.md) handles the JSON serializing and deserialization for us. However, for testing, we need to use `Marshal` and `Unmarshal`.
+
+[Here](https://github.com/ahsu1230/mathnavigatorSite/blob/master/orion/pkg/tests_integration/setup_test.go#L106) is an example of a helper function that serializes structs into JSON:
+
+```
+func createJsonBody(v interface{}) io.Reader {
+	marshal, _ := json.Marshal(v)
+	return bytes.NewBuffer(marshal)
+}
+```
```
ahsu1230, line 34: Also talk about the json tags in our domain structs. You know how we do the whole `program_id "json: programId..."`

Then talk about what omitempty is and the other json tag "-"

resources/backend/json.md
```diff
@@ -0,0 +1,49 @@
...
+Here is an example:
+```go
+type Example struct {
+	Id        uint      `json:"id"`
+	Empty     string    `json:"empty,omitempty"`
+	CreatedAt time.Time `json:"-"`
+}
+```
+
+In this struct, when we serialize, the `Id` field will have a JSON field of `id`. The `omitempty` means that if `Empty` is empty, it will not show up in the JSON. `-` means that this field will be omitted from the JSON no matter what.
```
ahsu1230, line 36: Bullet point these 3 things. "The code above has the following implications:"
 - ...Id -> id
 - ... omitempty
 - ...-

PR #125, opened 2020-04-22
Final Documentation

mzhang2021:
This should be the last of all documentation.


orion/configs/config_local.yaml
```diff
@@ -0,0 +1,9 @@
+app:
```
ahsu1230, line 1: sorry big refactor time. Go ahead and remove this file from the PR

orion/go.sum
```diff
@@ -0,0 +1,197 @@
+cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
ahsu1230, line 1: sorry big refactor happened, go ahead and remove this file from this PR

PR #134, opened 2020-06-05
Family and People

austinhsu14:


constellations/orion/src/controllers/family_controller.go
```diff
@@ -0,0 +1,104 @@
...
+	offset := ParseParamInt(c.Query("offset"), 0)
+
+	userList, err := repos.FamilyRepo.SelectAll(search, pageSize, offset)
+	if err != nil {
+		c.Error(err)\
+		c.String(http.StatusInternalServerError, err.Error())
+	} else {
+		c.JSON(http.StatusOK, userList)
+	}
+}
```
ahsu1230, line 24: I think we can get rid of this for now. It makes sense to search by specific users, But not so much for a list of families.

constellations/orion/src/controllers/family_controller.go
```diff
@@ -0,0 +1,104 @@
...
+		c.Error(err)
+		c.String(http.StatusNotFound, err.Error())
+	} else {
+		c.JSON(http.StatusOK, &family)
+	}
+}
+
+func GetFamilyByPrimaryEmail(c *gin.Context) {
+	// Incoming parameters
+	primaryemail := ParseParamUint(c.Param("primaryEmail"))
```
ahsu1230, line 41: primaryEmail - use camelCase

constellations/orion/src/domains/family.go
```diff
@@ -0,0 +1,33 @@
...
+	Password   string       `json:"password" db:"password"`
+}
+
+// Class Methods
+
+func (family *Family) Validate() error {
+	// Retrieves the inputted values
+	password := family.Password
+
+	if password == "" || len(password) > 32 {
```
ahsu1230, line 29: hmm, let's just let the DB take care of password length. So if the column only wants to keep 256 bytes of characters, if the password size goes over, the db will signal an error.

ahsu1230: so remove the `len(password) >` part

constellations/orion/src/domains/people.go
```diff
@@ -0,0 +1,82 @@
...
+		if guardianId.Valid {
+			return errors.New("guardian cannot have a guardian id")
+		}
+	} else {
+		if guardianId.Uint == id || !guardianId.Valid {
+			return errors.New("invalid guardian id")
+		}
+	}
+
+	// // Family validation
```
ahsu1230, line 72: I think no need to validate familyId, the DB will take care of that for us

constellations/orion/src/repos/family_repo.go
```diff
@@ -0,0 +1,198 @@
...
+
+// Implements interface userRepoInterface
+type familyRepo struct {
+	db *sql.DB // golang native db connection
+}
+
+// Interface to implement
+type FamilyRepoInterface interface {
+	Initialize(db *sql.DB)
+	SelectAll(string, int, int) ([]domains.Family, error)
```
ahsu1230, line 21: I think for now, get rid of SelectAll() for Family.
Some tests will need to change because you're not selecting all families anymore

constellations/orion/src/domains/people.go
```diff
@@ -0,0 +1,82 @@
...
+	CreatedAt  time.Time    `json:"-" db:"created_at"`
+	UpdatedAt  time.Time    `json:"-" db:"updated_at"`
+	DeletedAt  sql.NullTime `json:"-" db:"deleted_at"`
+	FirstName  string       `json:"firstName" db:"first_name"`
+	LastName   string       `json:"lastName" db:"last_name"`
+	MiddleName NullString   `json:"middleName" db:"middle_name"`
+	Email      string       `json:"email"`
+	Phone      string       `json:"phone"`
+	IsGuardian bool         `json:"isGuardian" db:"is_guardian"`
+	GuardianId NullUint     `json:"guardianId" db:"guardian_id"`
```
ahsu1230, line 24: I realize we don't need guardianId anymore. It actually restricts us a lot.

We can support multiple guardians to a family if we just have "familyId" and "isGuardian"
So we can have 6 users with the same familyId, 2 of them could be Guardians because isGuardian is true for both of them. The Primary email will be dependent on the user managing the family account.

constellations/orion/src/repos/people_repo.go
```diff
@@ -0,0 +1,232 @@
...
+type userRepo struct {
+	db *sql.DB // golang native db connection
+}
+
+// Interface to implement
+type UserRepoInterface interface {
+	Initialize(db *sql.DB)
+	SelectAll(string, int, int) ([]domains.User, error)
+	SelectById(uint) (domains.User, error)
+	SelectByGuardianId(uint) ([]domains.User, error)
```
ahsu1230, line 23: Replace SelectByGuardianId with selectByFamilyId(), which will give us a list of users.

constellations/orion/src/repos/people_repo.go
```diff
@@ -0,0 +1,232 @@
...
+	Insert(domains.User) error
+	Update(uint, domains.User) error
+	Delete(uint) error
+}
+
+func (ur *userRepo) Initialize(db *sql.DB) {
+	ur.db = db
+}
+
+func (ur *userRepo) SelectAll(search string, pageSize, offset int) ([]domains.User, error) {
```
ahsu1230, line 33: Keep this selectAll for users! We'll need this.

constellations/orion/src/router/handler.go
```diff
@@ -90,5 +90,15 @@ func (h *Handler) SetupApiEndpoints() {
...
 	}
+
+	apiFamilies := h.Engine.Group("api/families")
+	{
+		api.Families.GET("/all", controllers.GetAllFamilies)
+		api.Families.POST("/create", controllers.CreateFamily)
+		api.Families.GET("/family/:id", controllers.GetFamilyById)
+		api.Families.POST("/family/:id", controllers.UpdateFamily)
+		api.Families.DELETE("/family/:id", controllers.DeleteFamily)
+		api.Families.GET("/family/:primaryEmail",controllers.GetFamilyByPrimaryEmail)
```
ahsu1230, line 101: oh hmm I think this will cause some problems later. I don't think you can have a url that looks like:
```
api/families/family/aaron@gmail.com
```
Instead, change the url to:
```
api/families/search
```
And have primaryEmail come in as part of the bodyJson in the controller part (like how create/update do it)

PR #146, opened 2020-06-17
create intern page (gemini user) 

srenaxu:
@ahsu1230


constellations/gemini-user/src/internship/internship.sass
```diff
@@ -0,0 +1,56 @@
...
+    font-family: $fontFamilyRoboto
+    width: 65%
+    margin-left: auto
+    margin-right: auto
+
+
+    & h1
+        color: $colorTurquoise
+        margin-top: 30px
+        margin-bottom: 30px
```
mythicat554, line 13: Take a look at this file: https://github.com/ahsu1230/mathnavigatorSite/blob/master/resources/frontend/css_ordering.md
It lays out the conventional ordering for CSS properties. So, in this case, margin-top and margin-bottom would come before color.

constellations/gemini-user/src/internship/internship.sass
```diff
@@ -0,0 +1,56 @@
+@import "../utils/constants.sass"
+
+#view-intern
+    font-family: $fontFamilyRoboto
+    width: 65%
+    margin-left: auto
+    margin-right: auto
```
mythicat554, line 7: Just to adhere more to CSS ordering, put font-family below margin-right (look at the comment I left at Line 13- I saw this afterwards oops)

constellations/gemini-user/src/internship/internship.sass
```diff
@@ -0,0 +1,56 @@
...
+        margin-right: auto
+
+        & caption
+            color: $colorTurquoise
+
+        & td, th
+            border: 1px solid
+            border-color: #dddddd
+            text-align: left
+            padding: 8px
```
mythicat554, line 51: text-align should come after padding

constellations/gemini-user/src/internship/internship.js
```diff
@@ -0,0 +1,116 @@
...
+                            Internship opportunities will be announced when they
+                            are available. When they are, student candidates
+                            must first pass a series of coding assessments and
+                            interviews in order to receive a position. This
+                            interview process reflects the same process that
+                            famous technology companies like Google, Facebook,
+                            Amazon, etc. have. We encourage all students to
+                            attempt these assessments to familiarize themselves
+                            with this interview process structure if they ever
+                            want to pursue a career at any of these companies.
```
mythicat554, line 65: I'm not sure if this would work... but maybe you could give this specific caption and image different class names so that you could assign separate css to make it fit better? Or maybe @ahsu1230 has a way to modify your current css so that it fits nicely.

srenaxu: since these images and captions are just placeholders, i wanted to style it in a way that no matter what we insert into the placeholders, the captions will be aligned and the images will be the same size and aligned. i thought about styling each container separately but if we insert something other than the placeholder, then it'll probably look messy again. (fyi i gave the container and the images border colors so it's easier to see the container height etc)

constellations/gemini-user/src/internship/internship.sass
ahsu1230: Can you link the CSS ordering doc that we have in the `resources` folder plz? :) I think you made it.

constellations/gemini-user/src/internship/internship.sass
```diff
@@ -0,0 +1,56 @@
...
+    & h1
+        color: $colorTurquoise
+        margin-top: 30px
+        margin-bottom: 30px
+
+    .container-main
+        flex-direction: row
+        display: flex
+        justify-content: space-between
+        align-items: center
```
ahsu1230, line 19: I think CSS ordering is important because it prevents very simple mistakes and it's easy to find the properties you want if they're always consistent across all files.

I would order the following like so:
```
display
flex-direction
justify-content
align-items
```

constellations/gemini-user/src/internship/internship.sass
```diff
@@ -0,0 +1,56 @@
...
+
+        border: 1px solid
```
ahsu1230, line 21: you can just do `border: 1px solid $colorOrange`

constellations/gemini-user/src/internship/internship.sass
```diff
@@ -0,0 +1,56 @@
...
+        border-color: $colorOrange
+
+
+        .child-image
+            flex-grow: 1
+            border: 1px solid
+            border-color: #colorBlack
+
+            .image
+                width: 90%
```
ahsu1230, line 31: why not 100%?

constellations/gemini-user/src/internship/internship.js
```diff
@@ -0,0 +1,116 @@
...
+                        <img class="image" src={ApCalcImgScr} />
+                    </div>
+                    <div class="child-caption">
+                        <p>
+                            Math Navigator occassionally offers Software
+                            Development Internship for ambitious students who
+                            would like to pursuse a college major or a career in
+                            software development.
+                        </p>
+                    </div>
```
ahsu1230, line 47: I think flex-items don't necessarily have to be `<div>` elements. They can very well be anything.
```
<div class="container-main">
    <img .../>
    <p>...</p>
</div>
```

constellations/gemini-user/src/internship/internship.js
ahsu1230: Good point cathy. For now, I this is fine because I don't have real text yet. They're not going to be big paragraphs anyway. Probably a sentence or two.

ahsu1230:
After taking a look and trying it manually, it looks very nice. The table is actually a nice touch. A few things:

1. I'm also looking for this page to be available to be navigated to through the Header. So as a user, I would like to scroll to "Accomplishments" and under it "Internships" should appear as a choice which I can then select to get to the page you currently have.

To do this, look under the `utils` folder for `links.js`. Under `SublinksAchieve` you should see a list of objects which you can add to in order to have your link show up!

2. I take back what I said about using `flex-grow`. Don't use it here this time. You can do something simpler like this:

```html
<div class="container">
                    <img src={...}/>
                    <p>...</p>
                </div>
```

And for SASS

```sass
.container
        display: flex
        flex-direction: row
        margin-bottom: ______ <- whatever looks nice here.

        & img
            width: 120px
            margin-right: 60px

        & p
            margin: 0 <- you need this because `<p>` is a block element that automatically has margins on it. It will kind of mess up our spacing so we'll just override it and make it 0.
```

notewio:
The page looks nice! Other than the couple of extra lines at the end of the SASS file the code looks good as well

ahsu1230:
> The page looks nice! Other than the couple of extra lines at the end of the SASS file the code looks good as well

If no other comments please "Approve this change". On the upper right corner in the "Files" tab, there should be a "Review changes" button. Under it, you can select "Approve changes".

constellations/docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
     image: mysql
     container_name: db-mysql
     ports:
-    - "3308:3306" # host port 3308 links to container port 3306
+    - "3306:3306" # host port 3306 links to container port 3306
```
ahsu1230, line 7: not sure how this changed. But can you change this back to `3308` please?

constellations/gemini-user/src/internship/internship.sass
```diff
@@ -0,0 +1,49 @@
...
+            border-color: #dddddd
+            padding: 8px
+            text-align: left       
+
+    #last-paragraph
+        margin-bottom: 300px 
+
+
+        
+
```
ahsu1230, line 49: nit: a couple of extra lines at end of sass file

PR #148, opened 2020-06-17
Achieve order

mythicat554:

PR #150, opened 2020-06-17
added column order to achievements

mythicat554:
this PR used to be PR#148 (https://github.com/ahsu1230/mathnavigatorSite/pull/148)


constellations/orion/src/repos/achieve_repo.go
```diff
@@ -170,8 +174,9 @@ func (ar *achieveRepo) Insert(achieve domains.Achieve) error {
 		"created_at, " +
 		"updated_at, " +
 		"year, " +
-		"message" +
-		") VALUES (?, ?, ?, ?)"
+		"message, " +
+		"`order` " +
```
ahsu1230, line 178: why does order here have `?

constellations/orion/src/repos/achieve_repo.go
```diff
@@ -195,7 +201,8 @@ func (ar *achieveRepo) Update(id uint, achieve domains.Achieve) error {
 	statement := "UPDATE achievements SET " +
 		"updated_at=?, " +
 		"year=?, " +
-		"message=? " +
+		"message=?, " +
+		"`order`=? " +
```
ahsu1230, line 205: same

constellations/orion/src/repos/achieve_repo.go
mythicat554: apparently "order" is a keyword in mysql, so I had to add backticks to make it work

constellations/orion/src/repos/achieve_repo.go
```diff
@@ -105,7 +107,7 @@ func (ar *achieveRepo) SelectAllUnpublished() ([]domains.Achieve, error) {
 func (ar *achieveRepo) SelectAllGroupedByYear() ([]domains.AchieveYearGroup, error) {
 	results := make([]domains.AchieveYearGroup, 0)
 
-	stmt, err := ar.db.Prepare("SELECT * FROM achievements ORDER BY year DESC")
+	stmt, err := ar.db.Prepare("SELECT * FROM achievements ORDER BY year DESC, `order` ASC")
```
ahsu1230, line 110: I don't think you need ``order``, do you?

ahsu1230: i mean the `

constellations/orion/src/repos/achieve_repo.go
ahsu1230: oh interesting.

constellations/orion/src/repos/achieve_repo.go
```diff
@@ -105,7 +107,7 @@ func (ar *achieveRepo) SelectAllUnpublished() ([]domains.Achieve, error) {
 func (ar *achieveRepo) SelectAllGroupedByYear() ([]domains.AchieveYearGroup, error) {
 	results := make([]domains.AchieveYearGroup, 0)
 
-	stmt, err := ar.db.Prepare("SELECT * FROM achievements ORDER BY year DESC")
+	stmt, err := ar.db.Prepare("SELECT * FROM achievements ORDER BY year DESC, position ASC")
```
tonywu315, line 110: This line returns the achievements in the correct order in `MySQL`, but it seems that when `rows, err := stmt.Query()` is run, this order is disregarded and the first row in `rows` just becomes the first row that is added. @ahsu1230 @mythicat554

ahsu1230:
Taking a look at this and seeing where the problem is. The code looks right.

constellations/orion/src/repos/migrations/000017_add_position_achievements.down.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE achievements
```
ahsu1230, line 1: 17 is now taken up by the google classroom code. I think this should be file 19 now. Same with the `.up.sql` file.

constellations/orion/src/repos/achieve_repo_test.go
```diff
@@ -115,6 +115,17 @@ func TestSelectAllGroupedByYear(t *testing.T) {
...
+		).
+		AddRow(
+			3,
+			now,
+			now,
+			domains.NullTime{},
+			domains.NullTime{},
+			2021,
+			"800 on SAT Math",
+			1,
```
ahsu1230, line 128: as described from a DM,

when you're in `repo_test` there isn't any database logic. The `rows` variable here is supposed to represent whatever you want the database to return. So in the repo_test, there's no re-ordering of any sorts done (no database to do that). Instead, we have to specify the database is going to return *this*.

PR #153, opened 2020-06-18
Adding price_per_session

ChujiaGuo:


constellations/orion/src/repos/migrations/000021_add_price_per_session_at_classes.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE classes
+    ADD COLUMN price_per_session INT UNSIGNED;
```
ahsu1230, line 2: nice! looks good. By default, adding a column with any type can have a NULL value.
You can NOT allow null values by adding `NOT NULL` and/or `default 0`.

constellations/orion/src/tests_integration/class_integration_test.go
```diff
@@ -197,6 +197,7 @@ func createClass(id int) domains.Class {
 			EndDate:         later1,
 			GoogleClassCode: domains.NewNullString("ab12cd34"),
 			FullState:       0,
+			PricePerSession: 50,
```
ahsu1230, line 200: Make this NullUInt and make sure to test null values too

ChujiaGuo: Alright, should be done now.

PR #154, opened 2020-06-18
Adding price_lump

ChujiaGuo:


constellations/orion/src/domains/class.go
```diff
@@ -23,6 +23,7 @@ type Class struct {
 	StartDate       time.Time  `json:"startDate" db:"start_date"`
 	EndDate         time.Time  `json:"endDate" db:"end_date"`
 	GoogleClassCode NullString `json:"googleClassCode" db:"google_class_code"`
+	PriceLump       uint       `json:"priceLump" db:"price_lump"`
```
ahsu1230, line 26: I'm not sure if I mentioned this, apologies if I didn't. PriceLump is optional and can be NULL. So PriceLump type should be NullUint.

constellations/orion/src/repos/migrations/000022_add_price_lump_at_classes.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE classes
+    ADD COLUMN price_lump INT UNSIGNED;
```
ahsu1230, line 2: This is correct. If a column does NOT have `NOT NULL` or `DEFAULT 0` it means it can be NULL. So we just have to change the go code to support nullable values.

constellations/orion/src/domains/class.go
```diff
@@ -25,6 +25,7 @@ type Class struct {
 	GoogleClassCode NullString `json:"googleClassCode" db:"google_class_code"`
 	FullState       int        `json:"fullState" db:"full_state"`
 	PricePerSession NullUint   `json:"pricePerSession" db:"price_per_session"`
+	PriceLump       NullUint   `json:"priceLump" db:"price_lump"`
 }
 
 // Class Methods
```
ahsu1230, line 31: Can you add a validation + validationTest?

Classes should either have a PricePerSession or PriceLump.
So in other words, a valid class should have a non-null, non-zero PricePerSession OR a non-null PriceLump

ChujiaGuo: can there be both?

ahsu1230: for now, no.

ChujiaGuo: so either or then

ahsu1230:
@ChujiaGuo how are things coming along here? Are you having any trouble or bumping into any problems?

ChujiaGuo:
uh no, i didn't have access to my computer today, i'm currently doing my
final tests then i'll push to github

PR #159, opened 2020-06-23
Change location and program from styl to sass

DanielLiu14:


constellations/gemini-admin/src/location/location.sass
```diff
@@ -1,26 +1,26 @@
...
     position: relative
     width: 100%
     height: 400px
-    padding: 0 spacing2x
+    padding: 0 $spacing2x
     box-sizing: border-box
-    color: colorBlackDark
+    color: $colorBlackDark
 
     & h1
```
ahsu1230, line 11: you know what, let's make it consistent.

Anytime you see:

```
& h1
    ...
```

Just replace the `& h1` with `h1`. With the ampersand.
& selects the parent which is redundant if it's already indented

constellations/gemini-admin/src/location/locationEdit.sass
```diff
@@ -0,0 +1,41 @@
...
+
+    & button
+        padding: $spacing2x $spacing4x
+        border: none
+        border-radius: 6px
+        outline: none
+        font-size: $fontSize0
+        cursor: pointer
+
+        &:hover
```
ahsu1230, line 31: this is the only exception. We need `&:hover` here in order to select the hover state of the parent.

ahsu1230:
Unfortunately, looks like this PR is too stale and it's blocking other PRs #171. Let's close this PR and put up another change that changes location from styl to sass ASAP please.

PR #161, opened 2020-06-24
Admin announcements on home page & styl to sass

notewio:
@ahsu1230 @DanielLiu14 
The post request doesn't really work - the API call sometimes returns the error "announcement was not updated", and when it doesn't do that, nothing in the database changes


ahsu1230:
First off, orion has had a couple of changes since onboarding so you'll have to re spin it up. Use
```
docker-compose down
docker-compose build
docker-compose up -d
```

After you've reset orion, try it again.
From my end, when trying things out, it works when you select one of the checkboxes. And when you select one, it actually correctly saves because when you refresh the page, the correct announcement is still saved to be on the home page.

However, I get the error message you described when unchecking a checked announcement

ahsu1230:
You also have formatting issues and a couple of places with missing semi-colon. It may not be the reason for your bug, but I advise first fixing those so you can focus on logic instead of silly mistakes :)

constellations/gemini-admin/src/announce/announce.js
```diff
@@ -20,11 +20,51 @@ export class AnnouncePage extends React.Component {
...
+        let current = this.state.list.findIndex(
+            (announcement) => announcement.id == e.target.id
+        );
+
+        this.state.list[current].onHomePage = true;
+        if (previous >= 0) {
+            this.state.list[previous].onHomePage = false;
+        }
+
+        this.setState(this.state);
```
ahsu1230, line 42: this statement here is a bit odd. It seems like you are setting state to be what it already is. But I Also see why you need a statement like this because we are dealing with nested objects here (a list of objects and only one property of that field needs changing).

Here's what I recommend and to make this more "React-friendly".
```javascript
let list = this.state.list;

...change some stuff, edit variable list (NOT this.state.list)...
// maybe even this:
const finalList = ...;

this.setState({list: finalList});
```

It's not good practice to modify `this.state.list` because it gives the impression that values in it can change. Instead, save it into a new variable, modify that, and then replace the old one using `this.setState()`.

Hope that makes sense!

constellations/gemini-admin/src/announce/announce.sass
```diff
@@ -0,0 +1,71 @@
...
+
+    & .announce-list-row
+        display: block
+        position: relative
+        padding: $spacing2x 0
+        border-bottom: 1px solid $colorGrayLight
+        
+        display: flex
+        flex-direction: row
+        align-items: center
```
ahsu1230, line 29: this has display properties of both block and flex. If it's only display: flex, make sure the CSS properties go on the top for proper CSS ordering

constellations/gemini-admin/src/announce/announce.sass
```diff
@@ -0,0 +1,71 @@
...
+                    &.scheduled
+                        font-style: italic
+                        color: $colorGrayMed
+                    &.published
+                        color: $colorTurquoise
+
+                &.li-large
+                    width: $spacing56x
+                    margin-right: $spacing4x
+                    flex-grow: 1
```
ahsu1230, line 50: flex stuff should be first in the list of properties.

constellations/gemini-admin/src/announce/announce.sass
```diff
@@ -0,0 +1,71 @@
...
+                &.li-small
+                    width: $spacing12x
+
+                & div
+                    line-height: $lineHeight1
+
+        & a
+            //position: absolute
+            //top: 50%
+            //transform: translateY(-50%)
```
ahsu1230, line 61: if no longer needed, can remove.
Can remove position: relative from parent above as well.

constellations/gemini-admin/src/announce/announceEdit.sass
```diff
@@ -0,0 +1,50 @@
...
+
+  & .buttons
+    margin-top: $spacing2x
+    margin-bottom: $spacing8x
+
+    & button
+      outline: none
+      border: none
+      border-radius: 6px
+      padding: $spacing2x $spacing4x
```
ahsu1230, line 36: usually margin/padding comes before outline & border

PR #162, opened 2020-06-25
Sessions page

tonywu315:


constellations/gemini-admin/src/index.js
```diff
@@ -10,32 +10,44 @@ import { ProgramPage } from "./programs/program.js";
...
-const Achieve = () => <AchievePage />;
-const AchieveEdit = () => <AchieveEditPage />;
-const AchieveEditMatch = ({ match }) => (
-    <AchieveEditPage Id={match.params.Id} />
-);
-const Announce = () => <AnnouncePage />;
-const AnnounceEdit = () => <AnnounceEditPage />;
-const AnnounceEditMatch = ({ match }) => (
-    <AnnounceEditPage announceId={match.params.announceId} />
-);
```
mythicat554, line 31: There isn't anything you have to change, but I'm just wondering why you chose to put the files in this particular order?

constellations/gemini-admin/src/session/sessionAdd.js
```diff
@@ -0,0 +1,175 @@
...
+
+    render = () => {
+        return (
+            <section id="add-sessions">
+                <h3>Add Sessions</h3>
+
+                <div className="container">
+                    <div id="date-inputs">
+                        <div className="row">
+                            <div className="item">
```
mythicat554, line 97: Is there any way to use less divs here?

constellations/gemini-admin/src/index.js
tonywu315: I just put them in the same order as the tabs.

constellations/gemini-admin/src/session/sessionAdd.js
tonywu315: Not really. The item div is necessary to keep the label and the date picker together. The row is a flexbox. The `date-inputs` holds the 4 input things together and the `container` is another flexbox that helps with formatting the button next to the inputs.

constellations/gemini-admin/src/index.js
```diff
@@ -71,11 +77,22 @@ class App extends React.Component {
                 <Switch>
                     <Route path="/" exact component={Home} />
                     <Route
-                        path="/program/:programId/edit"
+                        path="/programs/:programId/edit"
```
ahsu1230, line 80: oh nice. Thanks

constellations/gemini-admin/src/session/sessionEdit.sass
```diff
@@ -0,0 +1,45 @@
...
+        align-items: center
+
+        h4
+            margin: 0
+
+        input
+            width: $spacing2x
+            margin: 0 $spacing2x
+    
+    buttons
```
ahsu1230, line 40: did you mean `button` or a class `.buttons`?

ahsu1230:
Oooh this is coming along nicely! Here are a few things

1) When I select a class from the dropdown menu (on the sessions page), the sessions you show must be the sessions for that class. So every time I select a new classId, the session rows need to change. So you will need to call the GET sessions api on every change (use `onChange` on the <select>)

2) Similarly, when you add sessions, you are only adding sessions to the selected classId. I think you might be doing that, but it's hard to tell which sessions belong to which classId.

ahsu1230:
Up to you if you want to do this now or in separate PR. But you can also remove some code in the classes flow. We only want to manage sessions on this page, so you can remove the sessions stuff from the classEdit page. Remove anything after the "Display Time" section.

tonywu315:
I will create another PR for the classes page to change to sass, use flexbox, and remove the sessions part.

constellations/gemini-admin/src/session/session.js
```diff
@@ -0,0 +1,106 @@
...
+        API.post("api/sessions/create", sessionsJSON)
+            .then(() => {
+                console.log("Successfully created sessions!");
+            })
+            .catch((err) => {
+                window.alert("Could not create sessions: " + err);
+            })
+            .finally(() => {
+                this.fetchSessionData(this.state.classId);
+            });
```
ahsu1230, line 66: very nice!

constellations/gemini-admin/src/session/session.js
```diff
@@ -0,0 +1,106 @@
...
+                classId: classId,
+            });
+            this.fetchSessionData(this.state.classId);
+        });
+    };
+
+    fetchSessionData = (classId) => {
+        API.get("api/sessions/class/" + classId).then((res) => {
+            var newSessions = res.data;
+            newSessions = _.sortBy(newSessions, ["startsAt"]);
```
ahsu1230, line 30: I'm surprised _.sortBy works. Usually we're supposed to import lodash functions individually. Maybe i botched the library importing... :0

constellations/gemini-admin/src/session/session.sass
```diff
@@ -0,0 +1,19 @@
+@import "../_.constants.sass"
+
+#select-class
+    display: flex
+    flex-flow: row wrap
+    align-items: center
+    max-width: $spacing96x
+    margin: auto
+    margin-top: $spacing4x
```
ahsu1230, line 9: I'd just combine these into:
`margin: $spacing4x auto 0`

constellations/gemini-admin/src/session/session.js
```diff
@@ -0,0 +1,106 @@
...
+                        id="dropdown"
+                        onChange={(e) => this.onChangeSelect(e)}>
+                        {classOptions}
+                    </select>
+                </section>
+                <SessionAdd
+                    classId={this.state.classId}
+                    addSessions={this.addSessions}
+                />
+                {sessionsList}
```
ahsu1230, line 102: Good job splitting this into two sections. I think it's def the right call

constellations/gemini-admin/src/session/sessionAdd.sass
```diff
@@ -0,0 +1,48 @@
...
+        margin: $spacing1x 0
+        padding: 0
+        font-size: $fontSize0
+
+    #repeat
+        width: $spacing24x
+
+    #btn-add-sessions
+        margin: $spacing2x
+        margin-right: 0
```
ahsu1230, line 45: `margin: $spacing2x 0`

spacing2x for top & bottom
0 for left & right

constellations/gemini-admin/src/session/sessionList.sass
```diff
@@ -0,0 +1,36 @@
+@import "../_.constants.sass"
+
+#session-rows
+    max-width: $spacing96x
+    margin: auto
+    margin-top: $spacing4x
```
ahsu1230, line 6: margin: $spacing4x auto 0

constellations/gemini-admin/src/session/sessionList.js
```diff
@@ -0,0 +1,64 @@
...
+                    <span className="large-column">{session.notes}</span>
+                    <span className="edit">
+                        <Link
+                            to={
+                                "/sessions/" +
+                                classId +
+                                "/" +
+                                session.id +
+                                "/edit"
+                            }>
```
ahsu1230, line 43: This part is a bit hard to read. Maybe save the value into a link?
```
const url = "sessions/" + classId + "/" + session.id + ...;

<Link to={url}/>
```

constellations/gemini-admin/src/session/sessionEdit.js
```diff
@@ -0,0 +1,275 @@
...
+                            }}
+                        />
+                    </div>
+                </div>
+
+                <div id="cancel-toggle" className="item">
+                    <h4>Canceled</h4>
+                    <input
+                        type="checkbox"
+                        value={this.state.canceled}
```
ahsu1230, line 250: I think this is not the correct way to set a checkbox as checked or not.
You'll need something like this: https://stackoverflow.com/questions/39120007/setting-a-checkbox-check-property-in-react

You have to update the "checked" attribute of <input> in order to have it checked by default

constellations/gemini-admin/src/localStorage.js
```diff
@@ -0,0 +1,11 @@
+const STORAGE_CURRENT_CLASS = "storage_current_class";
+
+export const getCurrentClassId = () => {
+    const classId = window.localStorage.getItem(STORAGE_CURRENT_CLASS);
+    return JSON.parse(classId);
+};
+
+export const updateCurrentClassId = (c) => {
```
ahsu1230, line 8: setCurrentClassId() we can mirror the localStorage API here (get & set)

constellations/gemini-admin/src/session/session.js
```diff
@@ -0,0 +1,117 @@
...
+
+export class SessionPage extends React.Component {
+    state = {
+        classes: [],
+        classId: "",
+        sessions: [],
+    };
+
+    componentDidMount = () => {
+        console.log("WHAT");
```
ahsu1230, line 17: remove :)

tonywu315: oops lol

PR #163, opened 2020-06-25
update home dashboard

srenaxu:
@ahsu1230 @notewio  currently, the unpublished classes do not reflect which classes are actually unpublished. i created placeholders.


notewio:
you could try wrapping each section in divs and setting the margins on those, or maybe set different classes for the h3s that need different spacing

notewio:
looks good

constellations/gemini-admin/src/home/home.js
```diff
@@ -1,205 +1,103 @@
...
+
+                    <div class="class-section">
+                        <p class="classID">Class ID</p>
+                        <p class="edit"> </p>
+                    </div>
+                    <div class="unpublished-classes">
+                        <p class="classID">ap_calculus_spring_2020</p>
+                        <p class="edit">
+                            {" "}
+                            <button class="editB">Edit</button>{" "}
```
ahsu1230, line 31: I think you don't need the `{" "}`

constellations/gemini-admin/src/home/home.js
```diff
@@ -1,205 +1,103 @@
...
+                        </p>
+                    </div>
+                    <div class="unpublished-classes">
+                        <p class="classID">ap_java_summer_2020_class_a</p>
+                        <p class="edit">
+                            {" "}
+                            <button class="editB">Edit</button>{" "}
```
ahsu1230, line 38: I think you don't need the {" "}

constellations/gemini-admin/src/home/home.js
```diff
@@ -1,205 +1,103 @@
...
-        this.onClickPage = this.onClickPage.bind(this);
-        this.onClickPublishAll = this.onClickPublishAll(this);
-    }
-
-    componentDidMount() {
-        API.get("api/unpublished").then((res) => {
-            const unpublishedList = res.data;
-            this.setState({
-                programs: unpublishedList.programs,
-                classes: unpublishedList.classes,
```
ahsu1230, line 36: Right now looks like you're displaying fake data which is fine for "looks" purposes. The next step is to hook up actual unpublished classes to be displayed. So you're going to need this part here.

in a componentDidMount(), call API.get("api/unpublished") which is a Promise. When you receive data, setState of `classes` to `unpublishedList.classes`. You'll only need unpublished classes here.

From there, now that you `classes` is a list of objects, you should be able to render every class as a row. Reach out to people if you have questions :)

ahsu1230:
I think some internship stuff accidentally made it into this change as well.

PR #165, opened 2020-06-29
responsive intern page

srenaxu:
@ahsu1230 @notewio


constellations/gemini-user/src/internship/internship.sass
```diff
@@ -42,4 +44,40 @@
...
+            align-items: center
+            border: 0.5px solid $colorGrayLight
+
+            & h4
+                margin: 2%
+                color: $colorTurquoiseLight
+                text-align: center
+
+            & p
+                margin: 5%
```
ahsu1230, line 79: I think to make the small view a lot simpler, just use `margin-bottom: ____px` to space grouped elements apart. So maybe a margin-bottom for each of the `<p>` elements. And if you need more spacing after the h4, use margins.

ahsu1230: You shouldn't have to use `display: flex` here.

notewio:
besides the things that aaron already mentioned, everything else looks good

srenaxu:
i got rid of the table in normal view. so its now flex-direction row in normal then column for mobile

constellations/gemini-user/src/internship/internship.js
```diff
@@ -78,59 +78,52 @@ export class InternshipPage extends React.Component {
...
-                    </tr>
-                    <tr>
-                        <td>Daniel L.</td>
-                        <td>Winston Churchill HS 2021</td>
-                        <td> </td>
-                    </tr>
-                </table>
+                <h4>Past Interns:</h4>
+                <div class="intern-main">
+                    <p class="name">Cathy Y.</p>{" "}
```
notewio, line 83: what does `{" "}` do here? when i tested it they didn't seem to make a difference in the page

srenaxu: not too sure. prettier added it

constellations/gemini-user/src/internship/internship.sass
```diff
@@ -20,26 +25,47 @@
...
-    #past-intern
-        border-collapse: collapse
-        border-spacing: 2px
-        width: 80%
-        margin: 30px auto
+    .intern-main
+        display: flex
+        flex-direction: row
+        & p
+            margin: 20px 20px 20px 0
```
ahsu1230, line 36: Just give this a these a bottom and right margin `margin: 0 20px 20px 0`. I think the interns section takes up too much room.

constellations/gemini-user/src/internship/internship.sass
```diff
@@ -20,26 +25,47 @@
...
-        & td, th
-            border: 1px solid
-            border-color: #dddddd
-            padding: 8px
-            text-align: left       
+            & img
+                margin-left: auto
+                margin-right: auto
+                margin-bottom: 0
+                width: 90%
```
ahsu1230, line 58: The images are bit too big in mobile view.
I think a width of 180px is good enough.
If you want to center using margin auto, add the following properties:
```
display: block
margin: 0 auto
width: _____
```

constellations/gemini-user/src/internship/internship.sass
```diff
@@ -11,6 +11,11 @@
         margin-bottom: 30px
         color: $colorTurquoise
 
+    & h4
+        margin-bottom: 20px
```
ahsu1230, line 15: When using hard-coded pixels, use the variables `spacing1x` found in `utils/constants.sass` instead. It keeps the spacing around the website more consistent.

Spacings like with margins or paddings should use `spacingxx`. Width and height of elements should use `propUnitxx`. Can you please replace all the values in this file with the appropriate variables?

PR #169, opened 2020-07-03
Start data

mythicat554:


ahsu1230:
Let's also make it so that the script doesn't stop making queries if one of them fails. So i would change `log.Fatalf(...)` to something else in the `postRequest` function

constellations/clients/orionFakeDataFiller/filler.go
```diff
@@ -35,6 +36,139 @@ func runFiller(hostAddress string) {
...
+
+	// Create announcements
+
+	body11 := strings.NewReader(`{
+								var now = time.Now().UTC()
+								"postedAt": now,
+								"author": "Author Name",
+								"message": "The Summer 2020 session of SAT Math will start on Tuesday, June 30.",
+								"onHomePage": true
+							}`)
```
ahsu1230, line 123: I see what you're trying to do.

You can try this: 
```
now := time.Now().UTC()
nowJson, _ := now.MarshalJSON()
bodyStr := fmt.Sprintf(`{
	"postedAt": %s,
	"author": "%s",
	"message": "%s",
	"onHomePage": true
}`, nowJson, "Author Name", "The Summer 2020 session of SAT Math")
body11 := strings.NewReader(bodyStr)
```

ahsu1230: The `%s` are replaced by arguments, specifically `nowJson`, the author name and the custom message.
Use this when you need to add a specific time or have specific arguments you want to insert.

mythicat554:
I wasn't sure how to get the sendPostRequest statements into helper functions since "hostAddress" is a parameter for the main function. 
Also, when I open Orion I get "[GIN-debug] [WARNING] Headers were already written. Wanted to override status code 400 with 200" for the two sessions, but it still shows up as a 200, so I'm not sure what's happening there.
@tonywu315

constellations/clients/orionFakeDataFiller/filler.go
```diff
@@ -20,29 +21,240 @@ import (
...
+		"onHomePage": %s
+	}`, nowJson, author, message, onHomePage)
+
+	return announceBody
+}
+
+func createClass(programId, semesterId, classKey, classId, locationId, times string) string {
+	now := time.Now().UTC()
+	nowJson, _ := now.MarshalJSON()
+	var later1 = now.Add(time.Hour * 24 * 30)
```
tonywu315, line 223: Since this is the only later variable, I think `later` is better.

constellations/clients/orionFakeDataFiller/filler.go
```diff
@@ -20,29 +21,240 @@ import (
...
+func createSession(classId, cancelled string) string {
+	now := time.Now().UTC()
+	nowJson, _ := now.MarshalJSON()
+
+	sessionBody := fmt.Sprintf(`{
+		"classId": "%s"
+		"startsAt": %s
+		"endsAt": %s
+		"cancelled": %s
+	}`, classId, nowJson, nowJson, cancelled)
```
tonywu315, line 248: `endsAt` should probably be after `startsAt`.

constellations/clients/orionFakeDataFiller/filler.go
```diff
@@ -20,29 +21,240 @@ import (
...
+	return sessionBody
 }
 
 func sendPostRequest(url string, body io.Reader) {
 	resp, err := http.Post(url, "application/json; charset=UTF-8", body)
 	if err != nil {
-		log.Fatalln(err)
+		log.Println("smth failed")
```
tonywu315, line 255: You could make this more descriptive by printing the error.

constellations/clients/orionFakeDataFiller/filler.go
```diff
@@ -20,29 +21,246 @@ import (
...
+	}`, classId, nowJson, laterJson, cancelled))
+	sendPostRequest(hostAddress+"/api/sessions/create", sessionBody)
+	return nil
 }
 
 func sendPostRequest(url string, body io.Reader) {
 	resp, err := http.Post(url, "application/json; charset=UTF-8", body)
 	if err != nil {
-		log.Fatalln(err)
+		log.Println("Post request was not fulfilled.")
```
ahsu1230, line 261: Can we display the `err` here as part of the PrintLn?

constellations/clients/orionFakeDataFiller/filler.go
```diff
@@ -20,29 +21,246 @@ import (
...
 	}
 	if resp.StatusCode != 200 && resp.StatusCode != 204 {
-		log.Fatalln("Response status was not successful.", resp)
+		log.Println("Response status was not successful.")
```
ahsu1230, line 264: Also print what the StatusCode was

constellations/clients/orionFakeDataFiller/filler.go
```diff
@@ -20,29 +21,246 @@ import (
...
+		"startsAt": %s
+		"endsAt": %s
+		"cancelled": %s
+	}`, classId, nowJson, laterJson, cancelled))
+	sendPostRequest(hostAddress+"/api/sessions/create", sessionBody)
+	return nil
 }
 
 func sendPostRequest(url string, body io.Reader) {
 	resp, err := http.Post(url, "application/json; charset=UTF-8", body)
```
ahsu1230, line 259: It would be great if we knew which request we're attempting and the "progress" of the script as we're sending so many requests. It would be useful to know which request succeeded and which failed. And if it failed, why.

So like you can have a simple print statement before the post like:
```
Attempting POST to ____________ ...
```
OR
Have a print statement in each of the `create___()` functions.
```
Creating program ap_calc...
Creating program ap_java...
...
```

And so on.

mythicat554:
I'm getting this message "[WARNING] Headers were already written. Wanted to override status code 400 with 200" for the two session post requests (they still end up returning as 200s when I check Orion). I looked over the controller files, but I can't figure out why the 400 code is being called in the first place @ahsu1230

ahsu1230:
> I'm getting this message "[WARNING] Headers were already written. Wanted to override status code 400 with 200" for the two session post requests (they still end up returning as 200s when I check Orion). I looked over the controller files, but I can't figure out why the 400 code is being called in the first place @ahsu1230

oh yeah I've seen this before. It's very weird. It's a known "bug" and I'll look into it, so don't worry about it for now. I think even if it appears, the session is still persisted to the database.

PR #170, opened 2020-07-03
Classes formatting

tonywu315:


constellations/gemini-admin/src/classes/classEdit.sass
```diff
@@ -0,0 +1,66 @@
...
+        button
+            outline: none
+            border: none
+            border-radius: 6px
+            padding: $spacing2x $spacing4x
+            font-size: $fontSize0
+            cursor: pointer
+
+        :hover
+            filter: brightness(1.1)
```
ahsu1230, line 57: I think this is supposed to be indented

constellations/gemini-admin/src/classes/class.js
```diff
@@ -73,43 +73,88 @@ export class ClassPage extends React.Component {
...
+        }
+
+        if (numUnpublished > 0) {
+            // Use the correct word
+            const firstWord = numUnpublished == 1 ? "class" : "classes";
+            const secondWord = numSelected == 1 ? "class" : "classes";
+            return (
+                <div id="publish">
+                    <p>
+                        You have {numUnpublished} unpublished {firstWord}.{" "}
```
ahsu1230, line 111: Do you need the `{" "}`?

tonywu315:
The buttons in the sessions had colors and formatting before, which I think it got from other formatting files. I'll just add them to the `sessionsEdits.sass` as well.

PR #172, opened 2020-07-05
changed family to account

austinhsu14:
changed all family to account


austinhsu14:
Editing the migration file here was acceptable because we haven't used the user-related stuff or the account related stuff on the front-end yet.

PR #173, opened 2020-07-06
Achievements & semesters flexbox

notewio:
@ahsu1230 @DanielLiu14


ahsu1230:
Looks great! Can you also add some breathing space above the Add button and below the last row?

PR #177, opened 2020-07-08
User afh

mythicat554:


mythicat554:
I wasn't sure what to test for in user_afh_valid_test.go... I was thinking of testing for negative numbers, but uints are >=0

constellations/orion/src/repos/migrations/000026_create_table_user_afh.up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE user_afh
+(
+    id          int unsigned NOT NULL AUTO_INCREMENT,
+    user_id     int          NOT NULL,
```
tonywu315, line 4: Should these 2 be unsigned too?

constellations/orion/src/domains/user_afh_valid_test.go
```diff
@@ -0,0 +1,17 @@
+package domains_test
```
tonywu315, line 1: Since these are uints, I don't think you really need to validate anything.

constellations/orion/src/domains/user_afh.go
```diff
@@ -0,0 +1,28 @@
+package domains
+
+import "errors"
+
+var TABLE_USERAFH = "user_afh"
+
+type UserAfh struct {
+	Id     uint `json:"id"`
+	UserId uint `json:"user_id"`
```
ahsu1230, line 9: I think json here should be `userId` and the db should be `user_id`. Look at other domains for examples.

constellations/orion/src/domains/user_afh.go
```diff
@@ -0,0 +1,28 @@
...
+	AfhId  uint `json:"afh_id"`
```
ahsu1230, line 10: same here. Make `afhId` for json, but `afh_id` for db

Json dictates what the field will look like when coming in and out of the controller.
DB dictates what the field looks like in the database.

ahsu1230: but the struct field names UserId and AfhId are fine.

constellations/orion/src/domains/user_afh.go
```diff
@@ -0,0 +1,28 @@
...
+	if userId < 0 {
+		return errors.New("invalid userId")
+	}
+
+	// AfhId validation
+	if afhId < 0 {
+		return errors.New("invalid afhId")
+	}
+
+	return nil
```
ahsu1230, line 27: yea agree with tony. Since there's not much to validate here, no need for a Validate() function for user_afh.

constellations/orion/src/domains/user_afh_valid_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"testing"
+
+	"github.com/ahsu1230/mathnavigatorSite/constellations/orion/src/domains"
+)
+
+func TestValidUserId(t *testing.T) {
```
ahsu1230, line 9: since nothing to validate, just this file not needed atm.

constellations/orion/src/repos/migrations/000026_create_table_user_afh.up.sql
ahsu1230: +1 unsigned int.

constellations/orion/src/repos/migrations/000026_create_table_user_afh.up.sql
```diff
@@ -0,0 +1,8 @@
...
+    afh_id      int          NOT NULL,
```
ahsu1230, line 5: unsigned int. Should match the column type of what you see in the ask for help table.

constellations/orion/src/repos/migrations/000026_create_table_user_afh.up.sql
```diff
@@ -0,0 +1,8 @@
...
+    PRIMARY KEY (id)
+) AUTO_INCREMENT = 1
+  DEFAULT CHARSET = UTF8MB4;
```
ahsu1230, line 8: `user_id` and `afh_id` should be foreign keys! Look at `classes` or `users` to see how to specify a foreign key. With foreign keys, they automatically ensure that "linked" entities exist.

For example, if I created a user_afh for user 432 and afh session 14, foreign keys ensure that user 432 exists in the user table and a afh session exists with id 14.

ahsu1230:
Nice job so far! It looks mostly good, we just need to add foreign keys to ensure fields from other tables are correctly linked together.

constellations/orion/src/tests_integration/user_afh_integration_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+// Create 3 UserAfhs. Then update one. Then get by afhId
+func Test_GetUserAfhsByAfhId(t *testing.T) {
+	createAllUserAfhs(t)
+
+	// Update
+	updatedUserAfh := createUserAfh(2, 1)
+	updatedBod := utils.CreateJsonBody(&updatedUserAfh)
+	recorder2 := utils.SendHttpRequest(t, http.MethodPost, "/api/userAfhs/userAfh/1", updatedBod)
+	assert.EqualValues(t, http.StatusOK, recorder2.Code)
```
mythicat554, line 42: I'm not sure why this is returning with "Error #01: sql: expected 3 arguments, got 4", especially since I pretty much did the same thing when I was testing the endpoints in my user_afh_controller_test.go file. Otherwise, the code works fine.

constellations/orion/src/router/handler.go
```diff
@@ -109,4 +109,14 @@ func (h *Handler) SetupApiEndpoints() {
 		apiAFH.DELETE("/afh/:id", controllers.DeleteAFH)
 	}
 	// apiAccounts := router.Group("api/accounts/")
+
+	apiUserAfh := h.Engine.Group("api/userAfhs")
```
ahsu1230, line 113: so turns out URLs are not case sensitive, to its the same as `api/userafhs`
As long as this is unique from the other above endpoints, I think it's ok to just keep this `api/userafhs`

constellations/orion/src/router/handler.go
```diff
@@ -109,4 +109,14 @@ func (h *Handler) SetupApiEndpoints() {
...
+	{
+		apiUserAfh.GET("/userAfhs/user/:userId", controllers.GetUserAfhByUserId)
```
ahsu1230, line 115: `/users/:userId`

constellations/orion/src/router/handler.go
```diff
@@ -109,4 +109,14 @@ func (h *Handler) SetupApiEndpoints() {
...
+		apiUserAfh.GET("/userAfhs/afh/:afhId", controllers.GetUserAfhByAfhId)
```
ahsu1230, line 116: `/afh/:afhId`

constellations/orion/src/router/handler.go
```diff
@@ -109,4 +109,14 @@ func (h *Handler) SetupApiEndpoints() {
...
+		apiUserAfh.GET("/userAfhs/user/:userId/afh/:afhId", controllers.GetUserAfhByBothIds)
```
ahsu1230, line 117: Nice!
Nit: `/users/:userId/afh/:afhId`

constellations/orion/src/router/handler.go
```diff
@@ -109,4 +109,14 @@ func (h *Handler) SetupApiEndpoints() {
 		apiAFH.DELETE("/afh/:id", controllers.DeleteAFH)
 	}
 	// apiAccounts := router.Group("api/accounts/")
```
ahsu1230, line 111: can you delete this please? we don't need anymore :)

constellations/orion/src/repos/user_afh_repo.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		return err
+	}
+	defer stmt.Close()
+
+	execResult, err := stmt.Exec(
+		userAfh.Id,
+		userAfh.UserId,
+		userAfh.AfhId,
+		id)
```
ahsu1230, line 142: found your problem! Looks like you're giving 4 arguments here, but there are only 3 `?`s above.

So this is where mocking is sometimes not useful. In this application, there are many moving parts. It would be very difficult to test all the moving parts at once! Instead, we want to test only certain components. But if components depend on each other (like A depends on B which depends on C...) how do we test just B? We can test B by *pretending* C works as expected. From there, the unit tests for B will test "Does B behave correctly if C works as expected?"

The _repo_test did not catch this because it's not running MySQL for real (that would require a MySQL server running and all this setup which is a hassle). The _repo_test is just making sure the correct MySQL query looks correct through pattern matching.

The _controller_test did not catch this because it depends on user_afh_repo. The _controller_test PRETENDS the repo is working correctly. If the repo works correctly, does the controller behave as expected?

So now you probably can see that there are some holes that these tests don't always catch. That's what integration tests are for and why you only bump into this problem in the integration tests.

So why not always use integration tests?
You could. But integration tests are a good chunk slower than unit tests because it requires network queries, MySQL servers & setups, etc. If it works for integration tests, it should *most likely* work correctly. However, if you had to test every single little detail with integration tests (like every possible bad inputs), the whole testing framework is going to take MUCH longer.

I should write a post in the resources folder about this lol but hopefully you have a better idea of what happened here. Super small mistake, but actually exposes a lot of nuance.

mythicat554: I see. So basically before, I was modifying the id too, but it should have only been modifying the userId and afhId by the id..?

ahsu1230: Ah yes, what you have now is correct. There are 3 question marks in lines 128-131, so you need to make sure your arguments then line up in order. The first one is userId, second is afhId and the third is the row Id of the `userAfh` object. In this case, you were actually fine using either `id` or `userAfh.Id` because they are the same (tho I prefer the former).

There are some domains (like programs) where you can actually update the `program_id`. So there's a sense of a new programId and an old programId. When updating, you have to be careful which one you want to use.

constellations/orion/src/controllers/controllers_util.go
```diff
@@ -36,3 +37,19 @@ func ParseParamInt(str string, preset int) int {
...
+func ParseParamUserId(c *gin.Context) uint {
+	id, err := strconv.ParseUint(c.Param("userId"), 10, 32)
+	if err != nil {
+		panic(err)
+	}
+	return uint(id)
+}
+
+func ParseParamAfhId(c *gin.Context) uint {
+	id, err := strconv.ParseUint(c.Param("afhId"), 0, 32)
```
ahsu1230, line 50: should this be 10 not 0?

PR #181, opened 2020-07-09
added dropdown for the full state of a class

DanielLiu14:
@ahsu1230 @srenaxu


ahsu1230:
Can you review this as well? @notewio

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -9,23 +9,44 @@ import { OkayModal } from "../modals/okayModal.js";
...
-
-        selectProgramId: "",
-        selectSemesterId: "",
-        selectLocationId: "",
-
-        programs: [],
-        semesters: [],
-        locations: [],
-        sessions: [],
-    };
```
ahsu1230, line 28: keep this format and add your additional variables. This "automatically" defines the constructor function and initializes the state variable. It's the more modern away to define state in ReactJs

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -9,23 +9,44 @@ import { OkayModal } from "../modals/okayModal.js";
...
+        this.onClickDelete = this.onClickDelete.bind(this);
+        this.onClickSave = this.onClickSave.bind(this);
+
+        this.onModalDeleteConfirm = this.onModalDeleteConfirm.bind(this);
+        this.onModalOkSaved = this.onModalOkSaved.bind(this);
+        this.onModalDismiss = this.onModalDismiss.bind(this);
+
+        this.onAddSessions = this.onAddSessions.bind(this);
+        this.onDeleteSession = this.onDeleteSession.bind(this);
+    }
```
ahsu1230, line 49: Once you keep the original format above, you can get rid of this constructor method

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -102,6 +123,14 @@ export class ClassEditPage extends React.Component {
         this.setState({ [value]: event.target.value });
     };
 
+    onChangeSelect = (e) => {
```
ahsu1230, line 126: maybe rename to something more specific like `onChangeFullState`

constellations/gemini-admin/src/classes/classEdit.sass
```diff
@@ -11,6 +11,10 @@
     h3
         margin: $spacing2x 0
 
+        .avalibility
```
ahsu1230, line 14: availability

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -273,8 +303,14 @@ export class ClassEditPage extends React.Component {
                             this.handleChange(e, "inputTimeString")
                         }
                     />
-                </div>
+                    <h3 className="avalibility">Class Avalibility</h3>
```
ahsu1230, line 306: Class Availability

constellations/gemini-admin/src/classes/classEdit.styl
```diff
@@ -0,0 +1,70 @@
+@import "../static.styl"
```
ahsu1230, line 1: we shouldn't be using styl anymore. Only sass.

constellations/gemini-admin/src/classes/classEdit.sass
```diff
@@ -11,6 +11,10 @@
...
+            padding-top: $spacing2x
+            border-top: 1px solid
```
ahsu1230, line 16: I believe border needs a color too:
`border-top: 1px solid _____`

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -102,6 +123,14 @@ export class ClassEditPage extends React.Component {
         this.setState({ [value]: event.target.value });
     };
 
+    onChangeSelect = (e) => {
+        const value = e.target.value;
+        console.log(value);
```
ahsu1230, line 128: when you've confirmed things work, please remove "debugging" print statements. :)

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -273,8 +303,14 @@ export class ClassEditPage extends React.Component {
...
                         }
                     />
-                </div>
+                    <h3 className="avalibility">Class Avalibility</h3>
 
+                    <select onChange={(e) => this.onChangeSelect(e)}>
+                        <option value="0">Normal</option>
+                        <option value="1">Almost Full</option>
+                        <option value="2">Full</option>
+                    </select>
```
notewio, line 312: I feel like it would make more sense to have this in the class information section rather than the schedule section

srenaxu:
looks fine!

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -112,6 +121,7 @@ export class ClassEditPage extends React.Component {
             locationId: this.state.selectLocationId,
             classKey: this.state.inputClassKey,
             times: this.state.inputTimeString,
+            classState: this.state.fullState,
```
ahsu1230, line 124: this should be fullState

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -25,6 +25,8 @@ export class ClassEditPage extends React.Component {
         semesters: [],
         locations: [],
         sessions: [],
+
+        fullState: "",
```
ahsu1230, line 29: Let's have fullState always be a number so it starts at 0.

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -273,8 +283,14 @@ export class ClassEditPage extends React.Component {
                             this.handleChange(e, "inputTimeString")
                         }
                     />
-                </div>
 
+                    <h4 className="availability">Class Availability</h4>
+                    <select onChange={(e) => this.onChangeFullState(e)}>
```
ahsu1230, line 288: You can do smoething like this:
```
<select value={this.state.fullState} onChange={...}/>
```
which will match the fullState value at all times. Since fullState is a number (as mentioned before), you may have to do string conversions here because the below options values are all strings of 0, 1, 2.

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -102,6 +104,13 @@ export class ClassEditPage extends React.Component {
         this.setState({ [value]: event.target.value });
     };
 
+    onChangeFullState = (e) => {
+        const value = e.target.value;
+        this.setState({
+            fullState: value,
```
ahsu1230, line 110: e.target.value returns a string, So we'll need to use `parseInt()` to convert this to an integer

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -102,6 +106,14 @@ export class ClassEditPage extends React.Component {
         this.setState({ [value]: event.target.value });
     };
 
+    onChangeFullState = (e) => {
+        const value = e.target.value;
+        value = parseInt(value);
+        this.setState({
+            fullState: value,
```
ahsu1230, line 113: Close! But looks like it didn't work because of this error which I found in the Console tab of Inspect Element. Do you need a refresher on Inspect? We scratched the surface in the web dev class, but I could've been more detailed with it.

Screen Shot 2020-07-15 at 1 19 41 AM

Looks like `value` is a constant so it can't be changed. You'll have to use a new variable or simply do this:
```
this.setState({
    fullState: parseInt(value)
});
```

ahsu1230:
Also, for future PRs, if you applied a comment/suggestion, could you please click on "resolve conversation"? That way, when a reviewer looks back, things are less cluttered and also easy to see what's been changed and what still needs to be addressed. Thanks!

PR #183, opened 2020-07-11
home dashboard

srenaxu:
@ahsu1230 @notewio


constellations/gemini-admin/src/home/home.sass
```diff
@@ -0,0 +1,64 @@
...
+    .class-section
+        display: inline-block
+        width: 100%
+        & p
+            margin: 0
+            border-bottom: 1px solid $colorGrayDark
+        & ul
+              list-style-type: none
+              padding: 0
+              margin: 0
```
notewio, line 51: inconsistent indentation

constellations/gemini-admin/src/home/home.js
```diff
@@ -1,205 +1,110 @@
...
-                            Locations
-                        </button>
-                        <button onClick={this.onClickAchievements}>
-                            Achievements
-                        </button>
-                        <button onClick={this.onClickSemesters}>
-                            Semesters
+                <h1>Administrator Dashboard</h1>
+
+                <div class="section">
```
ahsu1230, line 34: In React, we want to use `className` instead of `class` because it gives us the option to use variables for class names. So please change all `class=` in this file to `className=`

constellations/gemini-admin/src/home/home.js
```diff
@@ -1,205 +1,110 @@
...
-                    </ul>
-                    <h2> Registrations </h2>
-                    <ul>
-                        <button>New Users</button>
-                        <button>Questions</button>
-                        <button>Complaints</button>
-                    </ul>
+                    </div>
+                    <div class="class-section">
+                        <p>Class ID</p>
```
ahsu1230, line 44: Just replace this with a `<div>` and give it a className (i.e. `<div className="list-header-item"`) to add the bottom underline under it. `<div>` are block elements which mean they take up the whole row, which is why `<p>` elements work. `<span>` elements are inline elements which means they can be placed adjacent to other inline elements.

constellations/gemini-admin/src/home/home.js
```diff
@@ -1,205 +1,110 @@
...
-                        </button>
-                        <button onClick={this.onClickSemesters}>
-                            Semesters
+                <h1>Administrator Dashboard</h1>
+
+                <div class="section">
+                    <div class="container-class">
+                        <span>
+                            <h3>Unpublished Classes</h3>{" "}
+                        </span>
```
ahsu1230, line 38: Just make this a `<h3>` element (without the spans), but give the h3 a className and give it the CSS property `display: inline-block`

ahsu1230:
Awesome! Looks ready to merge! For future PRs, if you applied a comment/suggestion, could you please click on "resolve conversation"? That way, when a reviewer looks back, things are less cluttered and also easy to see what's been changed and what still needs to be addressed. Thanks!

PR #185, opened 2020-07-13
Transaction table

ChujiaGuo:
Creating transaction table


constellations/orion/src/repos/transaction_repo_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+	// Mock DB statements and execute
+	result := sqlmock.NewResult(1, 1)
+	mock.ExpectPrepare("^INSERT INTO transactions").
+		ExpectExec().
+		WithArgs(
+			sqlmock.AnyArg(),
+			sqlmock.AnyArg(),
+			100,
+			"pay_paypal",
```
ahsu1230, line 85: once you use turn these values into constants, you can access them via `domains.PAY_PAYPAL`.

Now we have a way to ensure people use the correct payment type and less likely to have typos that persist into the database.

constellations/orion/src/repos/transaction_repo_test.go
```diff
@@ -0,0 +1,196 @@
...
+			domains.NewNullString("note1"),
+			1,
+		).WillReturnResult(result)
+	transaction := domains.Transaction{
+		Amount:       100,
+		PaymentType:  "pay_paypal",
```
ahsu1230, line 91: domains.PAY_PAYPAL

ahsu1230: and other places in this file

constellations/orion/src/tests_integration/transactions_integration_test.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/stretchr/testify/assert"
+	"net/http"
+	"testing"
+)
+
+// Test: Create 3 Transactions and GetAll()
+func Test_CreateTransactions(t *testing.T) {
+	createAccounts(t)
+
+	trans1 := createTransaction(1, 100, "pay_paypal", "notes1", 1)
```
ahsu1230, line 16: domains.PAY_PAYPAL

constellations/orion/src/domains/transaction.go
```diff
@@ -0,0 +1,38 @@
+package domains
+
+import (
+	"errors"
+	"time"
+)
+
+var TABLE_TRANSACTIONS = "transactions"
+
```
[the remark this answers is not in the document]
ahsu1230: Here we use capital-case SNAKE_CASE because it's usually the way to signify that this is a static non-changing value that can be used in many places around the codebase

constellations/orion/src/controllers/transaction_controller_test.go
```diff
@@ -0,0 +1,235 @@
...
+				1,
+				100,
+				"pay_paypal",
+				"notes1",
+				1,
+			),
+			testUtils.CreateMockTransaction(
+				2,
+				200,
+				"pay_paypal",
```
ahsu1230, line 30: domains.PAY_PAYPAL for whole file

constellations/orion/src/router/handler.go
```diff
@@ -109,6 +109,15 @@ func (h *Handler) SetupApiEndpoints() {
 		apiAFH.DELETE("/afh/:id", controllers.DeleteAFH)
 	}
 
+	apiTransaction := h.Engine.Group("api/transactions")
+	{
+		apiTransaction.GET("/all", controllers.GetAllTransactions)
```
ahsu1230, line 114: Let's add another endpoint (GET) that grabs ALL payment types. It can be at: `api/transactions/types`

constellations/orion/src/controllers/transaction_controller.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"net/http"
+
+	"github.com/ahsu1230/mathnavigatorSite/constellations/orion/src/domains"
+	"github.com/ahsu1230/mathnavigatorSite/constellations/orion/src/repos"
+	"github.com/gin-gonic/gin"
+)
+
+func GetAllTransactions(c *gin.Context) {
```
ahsu1230, line 11: Can I get one more endpoint here? It's a pretty simple one: At GET `api/transactions/types`, can the user get a list of all types of PAYMENT_TYPES? So it should simply return a list of strings: `[pay_paypal, pay_charge, pay_cash, ...]`

This is so the front-end and back-end will always be in sync with what transaction types there can be. You will only need a controller_test for this, no need for an integration test since there's no repo stuff involved here.

constellations/orion/src/domains/transaction.go
```diff
@@ -7,6 +7,14 @@ import (
...
 var TABLE_TRANSACTIONS = "transactions"
 
+const (
+	PAY_CASH   = "pay_cash"
+	PAY_CHECK  = "pay_check"
+	PAY_PAYPAL = "pay_paypal"
+	REFUND     = "refund"
+	CHARGE     = "charge"
+)
+
```
ahsu1230, line 17: Add the list of payments here (since it can then be used in multiple places of the codebase)

const ALL_TRANSACTION_TYPES = [...]

constellations/orion/src/controllers/transaction_controller.go
```diff
@@ -31,6 +31,7 @@ func GetTransactionById(c *gin.Context) {
 	}
 }
 
+//Create transaction from JSON
 func CreateTransaction(c *gin.Context) {
 	//JSON
```
ahsu1230, line 36: remove this comment then

constellations/orion/src/controllers/transaction_controller.go
```diff
@@ -84,3 +85,8 @@ func DeleteTransaction(c *gin.Context) {
 		c.Status(http.StatusOK)
 	}
 }
+
+func GetAllPaymentTypes(c *gin.Context) {
+	paymentTypes := []string{domains.PAY_PAYPAL, domains.PAY_CASH, domains.PAY_CHECK, domains.CHARGE, domains.REFUND}
+	c.JSON(http.StatusOK, paymentTypes)
```
ahsu1230, line 91: if the list is defined in transaction.go,
You can then just refer to the list constant here:
c.JSON(http.StatusOK, domains.ALL_TRANSACTION_TYPES)

constellations/orion/src/controllers/transaction_controller.go
ahsu1230: Oh your above comment was fine. The `//JSON` comment here was too generic

ahsu1230:
Looks good to merge when you're ready!

PR #186, opened 2020-07-17
afh sessions signup

srenaxu:
@ahsu1230 @notewio


constellations/gemini-user/src/afh/afh.sass
```diff
@@ -0,0 +1,46 @@
...
+        padding: 14px 16px
+        transition: 0.3s
+        background-color: $colorWhite
+        font-size: 17px
+        color: $colorBlack
+
+    .tab button:hover
+        cursor: pointer
+        background-color: $colorGrayLight
+    .tab button:active
```
srenaxu, line 34: i'm having a bit of trouble here. when i select on a tab, i want the background color to change (will change it to underline later). however the background only turns turquoise when i'm holding down on the button. after, i let go the color is back to white. idk if you guys have any ideas @notewio @ahsu1230

constellations/gemini-user/src/afh/afh.sass
```diff
@@ -0,0 +1,46 @@
...
+        background-color: $colorTurquoise
+
```
srenaxu, line 36: i'm having a bit of trouble here. when i select on a tab, i want the background color to change (will change it to underline later). however the background only turns turquoise when i'm holding down on the button. after, i let go the color is back to white. idk if you guys have any ideas @notewio @ahsu1230

notewio: I think you have to do this with js, maybe instead of `.tab button:active` have something like `.tab button.active` and then set the classes on the buttons accordingly

constellations/gemini-user/src/afh/afh.sass
```diff
@@ -0,0 +1,64 @@
...
+        & h1, .description
+            text-align: center
+
+        & li
+            width: 90%
+            margin: $spacing1x auto
+
+        .tab button
+            font-size: 8px
+
```
srenaxu, line 16: the font size of the buttons isn't changing

tonywu315: line 52 should be `className` instead of `class`

constellations/gemini-user/src/afh/afh.js
```diff
@@ -0,0 +1,119 @@
...
+
+    openSubject = (subjectName) => {
+        this.setState({
+            currentTab: subjectName,
+        });
+    };
+
+    selectCheckbox = (title, onSelectSession) => {
+        return (
+            <input
```
srenaxu, line 29: wanted to see if anyone had any ideas regarding the checkbox for each individual session.

notewio: doing it this way is probably the best in terms of keeping things organized, but I think it has to be its own component rather than a function

constellations/gemini-user/src/afh/afh.js
```diff
@@ -0,0 +1,152 @@
...
+                    {showSessions}
+                </div>
+            </div>
+        );
+    }
+}
+
+class CheckboxInput extends React.Component {
+    state = {
+        isActive: "false",
```
ahsu1230, line 127: make these boolean values so `isActive: false`.

And similar below, you can compare values by using
`this.state.isActive == false` Or just simply `!this.state.isActive`

constellations/gemini-user/src/afh/afh.js
```diff
@@ -0,0 +1,152 @@
+"use strict";
+require("./afh.sass");
+import React from "react";
+import API from "../utils/api.js";
+import moment from "moment";
+
+export class AFHPage extends React.Component {
+    state = {
+        currentTab: "Math",
```
ahsu1230, line 9: Is there a way we can keep the value "behind the scenes" lower case `math` but still keep the user-facing display "Math", "English", etc.? Typically for inner values, you don't want to avoid having capitalized values because it could introduce discrepancies.

You could define a Javascript object above that looks like this:

const subjectDisplayNames = {
    math: "Math",
    english: "English",
    computer: "Computer Programming"
}

constellations/gemini-user/src/afh/afh.js
```diff
@@ -0,0 +1,152 @@
...
+                        className={
+                            this.state.currentTab == "Computer Programming"
+                                ? "active"
+                                : ""
+                        }
+                        onClick={() =>
+                            this.openSubject("Computer Programming")
+                        }>
+                        Computer Programming
+                    </button>
```
ahsu1230, line 95: These 3 buttons are all kind of the same. Just with a few values that are different which we can pass in as component props.

So maybe something like this:
```
<TabButton currentTab={this.state.currentTab}/>
```

When you pass in this.state.currentTab ("math", "english", or "computer"), With the subjectDisplayNames I mentioned above, you also have access to what to present to the user!
`subjectDisplayNames[this.props.currentTab]`

constellations/gemini-user/src/afh/afh.js
```diff
@@ -0,0 +1,152 @@
...
+                </div>
+
+                <div
+                    className={
+                        this.state.currentTab == "Computer Programming"
+                            ? "showTab"
+                            : "hide"
+                    }>
+                    {showSessions}
+                </div>
```
ahsu1230, line 119: I think you only need one of these. Not all 3.
That's because `showSessions` is already filtered by subject! So I don't think you need to do any hiding here. showSessions will always be filtered by the current subject

```
let currentSub = this.state.sessions.filter(
            (session) => session.subject == this.state.currentTab
        );

        let showSessions = currentSub.map((row, index) => {...
```

constellations/gemini-user/src/afh/afh.js
```diff
@@ -0,0 +1,152 @@
...
+        if (this.state.isActive == "false") {
+            this.setState({
+                isActive: "true",
+            });
+        } else {
+            this.setState({
+                isActive: "false",
+            });
+        }
+        return this.state.isActive;
```
ahsu1230, line 143: I don't think you need to return anything here. It doesn't look like you're doing anything with what you return?

constellations/gemini-user/src/afh/afh.js
```diff
@@ -0,0 +1,152 @@
...
+    }
+}
+
+class CheckboxInput extends React.Component {
+    state = {
+        isActive: "false",
+    };
+    selectCheckbox = (e) => {
+        return <input className="select" type="checkbox" />;
+    };
```
ahsu1230, line 131: don't use a function here and just put this into the return (...) part below.

constellations/gemini-user/src/afh/afh.js
```diff
@@ -0,0 +1,152 @@
...
+        }
+        return this.state.isActive;
+    };
+
+    render() {
+        const row = this.props.row;
+        const checkbox = this.selectCheckbox;
+
+        return <div className="checkbox">{checkbox}</div>;
+    }
```
ahsu1230, line 151: So overall, great job refactoring this to have the `isActive` state variable applied to the whole thing.
But go ahead and include the <input> part into here:

```
return (
    <div className=""/>
        <input ... />
    <div>
);
```

I think you were missing the parantheses before that may be why you were having troubel.

constellations/gemini-user/src/afh/afh.js
```diff
@@ -0,0 +1,126 @@
...
+                <h1>Ask for Help</h1>
+                <div className="description">
+                    We provide free sessions for students to ask for additional
+                    assistance on any of our program subjects. Please fill the
+                    form to let us know you are coming. You must be registered
+                    with one of our programs to attend.
+                </div>
+
+                <h1>Ask for Help Sessions by Subject</h1>
+                <div className="tab">
```
ahsu1230, line 54: Give this classname `tabs` convey this contains multiple tabs

constellations/gemini-user/src/afh/afh.js
```diff
@@ -0,0 +1,126 @@
...
+                    <TabButton
+                        currentTab={this.state.currentTab}
+                        openSubject={this.openSubject}
+                        subjectDisplayNames={this.subjectDisplayNames}
+                    />
```
ahsu1230, line 59: I would have 3 `TabButtons` looking like this:
```
<TabButton
          show={this.state.currentSubject == "math"}
          subject={"math"}
 />
<TabButton
          show={this.state.currentSubject == "english"}
          subject={"english"}
 />
<TabButton
          show={this.state.currentSubject == "computer"}
          subject={"computer"}
 />
```

constellations/gemini-user/src/afh/afh.js
```diff
@@ -0,0 +1,126 @@
...
+
+        let showSessions = currentSub.map((row, index) => {
+            return <AfhSessionRow key={index} row={row} />;
+        });
+
+        const subjectDisplayNames = {
+            math: "Math",
+            english: "English",
+            computer: "Computer Programming",
+        };
```
ahsu1230, line 41: this can be defined at the top of the file (above `export class AFHPage extends React.Component`). Treat this as a global variable

constellations/gemini-user/src/afh/afh.js
```diff
@@ -0,0 +1,126 @@
...
+
+class TabButton extends React.Component {
+    render() {
+        let currentTab = this.props.currentTab;
+        const subjectDisplayNames = this.props.subjectDisplayNames;
+
+        return (
+            <button
+                className={currentTab == "Math" ? "active" : ""}
+                onClick={() => this.openSubject("Math")}>
```
ahsu1230, line 76: this.openSubject(subject)

constellations/gemini-user/src/afh/afh.js
```diff
@@ -0,0 +1,126 @@
...
+                Math
```
ahsu1230, line 77: {{displayName}}

constellations/gemini-user/src/afh/afh.js
```diff
@@ -0,0 +1,126 @@
...
+                <div className="showTab">{showSessions}</div>
+            </div>
+        );
+    }
+}
+
+class TabButton extends React.Component {
+    render() {
+        let currentTab = this.props.currentTab;
+        const subjectDisplayNames = this.props.subjectDisplayNames;
```
ahsu1230, line 71: let show = this.props.show; // If show is true, show the tab
// if it is false, then hide the tab

let subject = this.props.subject;
let displayName = subjectDisplayNames[subject]; // <= this will grab the value from `subjectDisplayNames` given the key `subject`. For example, `subjectDisplayNames["computer"]` will give us "Computer Programming"

constellations/gemini-user/src/afh/afh.js
ahsu1230: The idea here is to use the state variable `currentSubject` (which can be any of lowercase math, english or computer) to decide which tab you want to show.
And given a subject, you can then figure out which display name you want to show.

constellations/gemini-user/src/afh/afh.js
```diff
@@ -0,0 +1,126 @@
+"use strict";
+require("./afh.sass");
+import React from "react";
+import API from "../utils/api.js";
+import moment from "moment";
+
+export class AFHPage extends React.Component {
+    state = {
+        currentTab: "Math",
```
ahsu1230, line 9: rename this currentSubject. This value is ALWAYS one of (math, english, or computer).
We will use it as a key to figure out which subject we are dealing with atm

ahsu1230:
Messaged you over Slack regarding how to think about all of this. Putting it here for context

> I think overall, you’r pretty close to getting it right. But I thought I would give you more info to explain what I think would help organize the thought process behind your code. Here, I’ll summarize main tips that I gave.
There are 3 tab components in the HTML each with a subject value associated to it. The subject will always be the lower-case name short name (math, english, computer). You will use this lower-case subject as the “key” to figure out which sessions and displayName you want to display to the user. Whenever we click on a tab, we will change this key to another lower-case subject value. We can keep track of which tab we are on using a state variable `this.state.subject` or `this.state.currentSubject`. At the end of the day, subject will always be one of (math, english, computer).
2.  What is subjectDisplayNames for? The subject key is a way for us to internally check which tab we want to show. We usually don’t want this “internal” value to be exposed to the user. That’s because let’s say one day we think a user prefers the term “Mathematics” or “Programming” etc., if these values were our internal values, we would have to change it in so many places (since it’s user often in your code). Instead, we use subjectDisplayNames to map our internal value to our user-facing value. (i.e. computer -> Computer Programming) using a Javascript Object (key-value storing).
3.  We will use this subject key to also filter what sessions we want to see. This you’ve correctly done with these lines. Using subject , we can filter out of our big list of AFH sessions, which we one we want to see based on the internal subject value.
        let currentSub = this.state.sessions.filter(
            (session) => session.subject == this.state.currentTab
        );
4.  So which tab to highlight? Very simple! We have 3 TabButtons, each have a default highlight. However, we want to give one TabButton a more special highlight. Which one? It’s whoever’s subject matches our this.state.subject variable (or this.state.currentSubject) whichever makes more sense to you.
So summary: Everything is going to revolve around this one state variable subject. This value will always be one of: math english computer. When you click on a tab, we simply switch values. From there, based on this one value, you can figure everything else out!

constellations/gemini-user/src/afh/afh.js
```diff
@@ -0,0 +1,131 @@
...
+    componentDidMount() {
+        API.get("api/askforhelp/all").then((res) => {
+            const afh = res.data;
+            this.setState({
+                sessions: afh,
+            });
+        });
+    }
+
+    openSubject = (subjectName) => {
```
ahsu1230, line 28: I think I prefer the name `changeSubject` here

constellations/gemini-user/src/afh/afh.sass
```diff
@@ -0,0 +1,86 @@
+@import "../utils/constants.sass"
+
+
```
ahsu1230, line 3: extra spaces here

constellations/gemini-user/src/afh/afh.sass
```diff
@@ -0,0 +1,86 @@
...
+    .sessions-list-active
+        display: flex
+        flex-direction: row
+        align-items: center
+        margin: $spacing4x auto
+        color: $colorTurquoiseLight
+
+        .sessions-checkbox
+            width: 20%
+
```
ahsu1230, line 33: one extra space here

constellations/gemini-user/src/afh/afh.sass
```diff
@@ -0,0 +1,83 @@
...
+    .description
+        margin: $spacing2x 0
+        color: $colorBlack
+
+    .select
+        flex: 0 0 $spacing10x
+        margin-right: $spacing2x
+
+    .tabs 
+        overflow: hidden
```
ahsu1230, line 41: is this still needed?

constellations/gemini-user/src/afh/afh.js
```diff
@@ -0,0 +1,163 @@
...
+                onClick={() => this.props.onChangeTab(subject)}>
+                {displayName}
+            </button>
+        );
+    }
+}
+
+class LocationAddress extends React.Component {
+    state = {
+        location: [],
```
ahsu1230, line 96: Sorry just change this to:
`location: {}`

PR #189, opened 2020-07-19
Added user afh and class functionality

tonywu315:


constellations/gemini-admin/src/user/userAFH.js
```diff
@@ -0,0 +1,180 @@
...
+
+    fetchAFHs = (afhIds) => {
+        let searchArray = new Set(afhIds);
+        API.get("api/askforhelp/all")
+            .then((res) => {
+                var userAFHs = [];
+                var afhs = [];
+                res.data.forEach((afh) => {
+                    if (searchArray.has(afh.id)) {
+                        userAFHs.push(afh);
```
mythicat554, line 62: Is afh here an afh object? Just wondering how that gets appended to an array that's for userAFHs.

tonywu315: Yeah its the afh object that contains the userId and the afhId

tonywu315: `userAFHs` is the afhs that the user is signed up for and `afhs` is all the others

ahsu1230:
If this is going to merge into the other feature branch, you can do so whenever you want. I need to do a final review on the feature branch anyway

PR #191, opened 2020-07-21
More input fields for admin class page

DanielLiu14:


DanielLiu14:
@ahsu1230 @srenaxu

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -231,6 +245,72 @@ export class ClassEditPage extends React.Component {
...
+                        value={this.state.paymentNotes}
+                        onChangeCallback={(e) =>
+                            this.handleChange(e, "paymentNotes")
+                        }
+                        validators={[
+                            {
+                                validate: (text) => text != "",
+                                message: "You must input a description",
+                            },
+                        ]}
```
ahsu1230, line 311: Since payment notes is an optional field and can contain almost anything, I don't think we need any validators. You can just delete this part

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -231,6 +245,72 @@ export class ClassEditPage extends React.Component {
...
+                        description="Enter price for pay per session attended (Either enter only in this field or only in the price lump field)"
+                        onChangeCallback={(e) =>
+                            this.handleIntegerChange(e, "pricePerSession")
+                        }
+                        validators={[
+                            {
+                                validate: (text) => text != "",
+                                message: "You must input a description",
+                            },
+                        ]}
```
ahsu1230, line 294: A class should either have `pricePerSession` filled or `priceLump` filled or neither. If both of these fields are filled, they should be valid numbers (i.e. can't input abc as a price).

Since we're checking for two conditions, `validators` can be an array of two validators. One validator per condition & message.
```
validators={[
    {
         validate: (input) => // check valid uninteger,
         message: "price must be a valid number"
    },
    {
        validate: (input) => // check that if this input is not "", this.state.priceLump should be a valid number
        message: "Only one of PriceLump or PricePerSession should be filled. Both cannot be filled"
    }
]}
```

Then do the same for the priceLump input

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -231,6 +245,72 @@ export class ClassEditPage extends React.Component {
...
+                        description="Enter the google classroom code"
+                        value={this.state.googleClassCode}
+                        onChangeCallback={(e) =>
+                            this.handleChange(e, "googleClassCode")
+                        }
+                        validators={[
+                            {
+                                validate: (text) => text != "",
+                                message: "You must input a description",
+                            },
```
ahsu1230, line 261: This field can be optional so this field is allowed to be empty. I think we don't need a validator here since the GoogleCode can be a mixture of alphanumerical characters.

srenaxu:
everything looks good to me!

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -231,6 +245,72 @@ export class ClassEditPage extends React.Component {
...
+                        isTextBox={true}
+                        required={false}
+                        description="Enter price for one time payment (Either enter only in this field or only in the price per session field)"
+                        onChangeCallback={(e) =>
+                            this.handleIntegerChange(e, "priceLump")
+                        }
+                        validators={[
+                            {
+                                validate: (input) => parseInt(input) != NaN,
+                                message: "price must be a valid number",
```
ahsu1230, line 270: Price

ahsu1230, line 270: I think all of these fields should be outside the the if clause (!isEdit). And should be below the fullState section. We may want to adjust prices or the google classroom code later

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -122,8 +132,12 @@ export class ClassEditPage extends React.Component {
             semesterId: this.state.selectSemesterId,
             locationId: this.state.selectLocationId,
             classKey: this.state.inputClassKey,
+            googleClassCode: this.state.googleClassCode,
+            paymentNotes: this.state.paymentNotes,
```
ahsu1230, line 136: move these fields below to be next to priceLump / pricePerSession.
Order hsould be fullState, googleClassCode, priceX, paymentNotes

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -16,6 +17,8 @@ export class ClassEditPage extends React.Component {
         oldClassId: "",
         inputClassKey: "",
         inputTimeString: "",
+        googleClassCode: "",
+        paymentNotes: "",
```
ahsu1230, line 21: reorder below

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -294,6 +308,72 @@ export class ClassEditPage extends React.Component {
...
+                                validate: this.state.pricePerSession != 0,
+                                message:
+                                    "Only one of PriceLump or PricePerSession should be filled. Both cannot be filled",
+                            },
+                        ]}
+                    />
+
+                    <InputText
+                        label="Price Per Session"
+                        isTextBox={true}
```
ahsu1230, line 347: should not be a textBox (just a single number for input). Textboxes are for large texts like paymentNotes.

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -294,6 +308,72 @@ export class ClassEditPage extends React.Component {
...
+                        description="Enter the google classroom code"
+                        value={this.state.googleClassCode}
+                        onChangeCallback={(e) =>
+                            this.handleChange(e, "googleClassCode")
+                        }
+                    />
+
+                    <InputText
+                        label="Price Lump"
+                        isTextBox={true}
```
ahsu1230, line 325: should not be a textBox (just a single number for input). Textboxes are for large texts like paymentNotes.

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -294,6 +308,72 @@ export class ClassEditPage extends React.Component {
                         <option value="1">Almost Full</option>
                         <option value="2">Full</option>
                     </select>
+
+                    <InputText
+                        label="Google Classroom Code"
+                        isTextBox={true}
```
ahsu1230, line 314: should not be a textBox (just a single number for input). Textboxes are for large texts like paymentNotes.

PR #193, opened 2020-07-24
sort by semester endpoint

mythicat554:


constellations/orion/src/controllers/classes_by_semester_controller_test.go
```diff
@@ -0,0 +1,80 @@
+package controllers_test
```
ahsu1230, line 1: Make your test & regular controller file names consistent.
The test is called `classes_by_semester_controller_test.go` but the regular controller is: `classes_by_semesterS_controller.go`

constellations/orion/src/controllers/classes_by_semester_controller_test.go
```diff
@@ -0,0 +1,80 @@
...
+		}, nil
+	}
+	repos.SemesterRepo = &testUtils.SemesterRepo
+
+	var now = time.Now().UTC()
+	var later1 = now.Add(time.Hour * 24 * 30)
+	testUtils.ClassRepo.MockSelectAll = func(publishedOnly bool) ([]domains.Class, error) {
+		return []domains.Class{
+			testUtils.CreateMockClass(
+				"prog1",
```
ahsu1230, line 54: I think the problem is here. `prog1` and `prog2` are "not found" programs because they are not in the `ProgramRepo.SelectAll()`, so you end up getting programs that are unavailable and I think the code exits out prematurely? That might be why you're getting the 400 error

constellations/orion/src/controllers/classes_by_semesters_controller.go
```diff
@@ -0,0 +1,118 @@
...
+				classSlice = append(classSlice, classes[i])
+			}
+		}
+		semesterClassMap[semesterObj.SemesterId] = classSlice // list of classes in specific semester
+
+		// create list of programIds in specific semester
+		var programList []string
+		for i := 0; i < len(classSlice); i++ {
+			if Find(programList, classSlice[i].ProgramId) != -1 {
+				continue
```
ahsu1230, line 60: hmm I think it might be better here to have an error! If we have something wrong with our data where things aren't expected, we should error hard - meaning we stop the endpoint and respond with a InternalServerError or something (soft means like a warning / ignore). 

In this case, a class's programId is NOT in the all programs list which is very unusual, so it's something to alert developers.

constellations/orion/src/controllers/classes_by_semesters_controller.go
```diff
@@ -0,0 +1,118 @@
...
+
+			// make ProgramClass struct
+			programClassStruct = updateProgramClass(programObj, finalClassList)
+			listProgramClass = append(listProgramClass, programClassStruct) // append to list of ProgramClasses
+			finalClassList = nil
+		}
+
+		// make ProgramClassesBySemester struct
+		classSemesterJson = updateProgramClassesBySemester(value, listProgramClass)
+		c.BindJSON(&classSemesterJson)
```
ahsu1230, line 85: I think I see what you're trying to do here. You're adding JSON objects to the list to create the final JSON.
But I don't think you need to do that, `c.JSON` below will recursively convert your Go struct into JSON meaning the outer struct and inner structs will all apply their JSON transformations.

So go ahead and just remove this line.

constellations/orion/src/controllers/classes_by_semesters_controller.go
```diff
@@ -0,0 +1,118 @@
...
+		// Find classes that have the same semester Id and append to semesterClassMap
+		classSlice := make([]domains.Class, 0)
+		for i := 0; i < len(classes); i++ {
+			if classes[i].SemesterId == semesterObj.SemesterId {
+				classSlice = append(classSlice, classes[i])
+			}
+		}
+		semesterClassMap[semesterObj.SemesterId] = classSlice // list of classes in specific semester
+
+		// create list of programIds in specific semester
```
ahsu1230, line 56: Capitalize first word of all comments i.e.
Create list of programIds in specific semester

constellations/orion/src/controllers/classes_by_semesters_controller.go
```diff
@@ -0,0 +1,118 @@
...
+		var programObj domains.Program
+
+		// Find classes that have the same semester Id and append to semesterClassMap
+		classSlice := make([]domains.Class, 0)
+		for i := 0; i < len(classes); i++ {
+			if classes[i].SemesterId == semesterObj.SemesterId {
+				classSlice = append(classSlice, classes[i])
+			}
+		}
+		semesterClassMap[semesterObj.SemesterId] = classSlice // list of classes in specific semester
```
ahsu1230, line 54: Better to have comments on separate line before the code you want to describe

constellations/orion/src/controllers/classes_by_semester_controller_test.go
```diff
@@ -0,0 +1,80 @@
...
+			),
+		}, nil
+	}
+	repos.ClassRepo = &testUtils.ClassRepo
+
+	// Create new HTTP request to endpoint
+	recorder := testUtils.SendHttpRequest(t, http.MethodGet, "/api/classesbysemesters/group", nil)
+
+	// Validate results
+	assert.EqualValues(t, http.StatusOK, recorder.Code)
```
ahsu1230, line 79: Make sure to also test content of this recorder as well.
View how other controller tests check the result of this HTTP request.
i.e. 
```
result[0].Semester.SemesterId should be equal to ____
result[0].ProgramClasses[0].Program.ProgramId should be _______
etc.
```

constellations/orion/src/controllers/classes_by_semester_controller_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"time"
+
+	"github.com/ahsu1230/mathnavigatorSite/constellations/orion/src/controllers/testUtils"
+	"github.com/ahsu1230/mathnavigatorSite/constellations/orion/src/domains"
+	"github.com/ahsu1230/mathnavigatorSite/constellations/orion/src/repos"
+	"github.com/stretchr/testify/assert"
+)
+
+// Test getting programs and classes by semester endpoint
+func TestGetClassesAndProgramsBySemester_Success(t *testing.T) {
```
ahsu1230, line 15: I think the best way is to really guarantee this endpoint works is to use a series of test testing different scenarios.
One could be 1 semester, 1 program, 1 class.
Then 1 semester, 1 program, 2 classes
Then 1 semester 2 programs, 2 classes and so on.

constellations/orion/src/controllers/classes_by_semester_controller_test.go
```diff
@@ -0,0 +1,80 @@
...
+				ProgramId:   "ap_java",
+				Name:        "AP Java",
+				Grade1:      2,
+				Grade2:      3,
+				Description: "Description2",
+				Featured:    1,
+			},
+		}, nil
+	}
+	repos.ProgramRepo = &testUtils.ProgramRepo
```
ahsu1230, line 39: Becuase we'll have multiple tests, this part is going to be repeated a lot.Is there a way we make this easier on ourselves and simply have a function that looks like this:

`mockPrograms(...)` which we call in every test. so it shrinks the amount of code per test?

constellations/orion/src/controllers/classes_by_semester_controller_test.go
mythicat554: I changed prog1 and prog2 to ap_calculus and ap_java (which are a part of the ProgramRepo.SelectAll()), but it's still giving me a 400 error

constellations/orion/src/controllers/classes_by_semesters_controller.go
mythicat554: The way I sorted the objects here was to go through a list of classes for each semester and create a list of programs... So basically here, if the programId was already in programList, I just continued the loop onto the next programId. 

I think if there was actually a class's programId that wasn't in the all programs list, it would go wrong where I map from the programId to the full program object at ```programObj = programMap[programList[i]]```. So I guess I'll create an if-then statement there to check if the programId is inside the programMap.

constellations/orion/src/controllers/classes_by_semester_controller_test.go
mythicat554: I think I'll add in the rest of the code after I figure out why it's still returning a 400

constellations/orion/src/controllers/classes_by_semesters_controller.go
```diff
@@ -0,0 +1,125 @@
...
+import (
+	"net/http"
+
+	"github.com/ahsu1230/mathnavigatorSite/constellations/orion/src/domains"
+	"github.com/ahsu1230/mathnavigatorSite/constellations/orion/src/repos"
+	"github.com/gin-gonic/gin"
+)
+
+func GetAllProgramsSemestersClasses(c *gin.Context) {
+	var classSemester domains.ProgramClassesBySemester
```
ahsu1230, line 12: I'm not a fan of having this variable here. It's only used in the for loop so this variable should be defined only inside the for loop.
If it is defined outside the for loop, it tells me that it might be used OUTside the for loop which it never is.

constellations/orion/src/controllers/classes_by_semesters_controller.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	for i := 0; i < len(classes); i++ {
+		classMap[classes[i].ClassId] = classes[i]
+	}
+
+	// Semester to class list map (maps semesterId to Class object)
+	semesterClassMap := make(map[string][]domains.Class)
+
+	// Loop over semesterIds
+	for i := 0; i < len(semesters); i++ {
```
ahsu1230, line 41: Here's a suggestion to make this for loop overall easier.

```
var listResults []domains.ProgramClassesBySemester
for _______________ {
    semesterObj := semesters[i]
    programClasses := createProgramClassesForSemester(semestersObj.SemesterId)
    programClassesBySemester := domains.ProgramClassesBySemester{
		Semester:       semester,
		ProgramClasses: programClasses,
	}
    listResults = append(listResults, programClassesBySemester)
}
```

Now it's super easy to see what's going on in this layer.
You are mapping over a list, grabbing the semester. Somewhere you create a bunch of programClasses and bundle it together with this class.

ahsu1230: From there, let's define the createProgramClassesForSemester(...) - there's some logic differences that I think would make your code work faster and neater.

```
func createProgramClassesForSemester(semesterId, ... // other stuff like the classMaps or programMaps probably) {
    // define list of []ProgramClass
    // find classes with semesterId

    // Step A. This can be its own function too!
    // create a map of programId -> list of classes
    // for each class found, 
             extract programId
             add to map (programId -> list of classes (with new class))

    // Step B. Using the map from step A This might be its own function too!
    define an empty list of []ProgramClass
    // for each (program) in (list of programs from original programRepo.SelectAll)
         // 1. get list of classes for this programId
         // 2. for each class in this list, 
                       extract class from ClassMap (classObj)
                       create a ProgramClass from programObject and classObject
                       append to a list of []ProgramClass
   return []ProgramClass
}
```

Something like that ^. I might've missed some details though. But I think changing your functionality to use the part I labeled in step A will help a lot.

The overall idea here is to not have too many indents in a function. We call this modularization. A lot of people say 80 character limit to each line of code in order to keep it readable. If half of that is indents, you're wasting a lot of space and it's hard to code, so you should refactor things out into a separate function - and have more of that complicated logic isolated into functions.

constellations/orion/src/controllers/classes_by_semesters_controller.go
```diff
@@ -0,0 +1,125 @@
...
+		Classes:    classes,
+	}
+}
+
+func updateProgramClassesBySemester(semester domains.Semester, programClasses []domains.ProgramClass) domains.ProgramClassesBySemester {
+	return domains.ProgramClassesBySemester{
+		Semester:       semester,
+		ProgramClasses: programClasses,
+	}
+}
```
ahsu1230, line 116: I think these are nice tries, but I'm hoping these functions have more logic in them. These just create new objects which would probably help with shorthand. But I'm hoping more for functions with like for-loops in them or complicated map-creating in them, you know?

constellations/orion/src/controllers/classes_by_semesters_controller.go
```diff
@@ -0,0 +1,125 @@
...
+
+func Find(slice []string, val string) int {
```
ahsu1230, line 118: This is a great example of refactoring out logic!

constellations/orion/src/controllers/classes_by_semesters_controller_test.go
```diff
@@ -0,0 +1,223 @@
...
+	assert.EqualValues(t, "program1", results[0].ProgramClasses[0].ProgramObj.ProgramId)
+	assert.EqualValues(t, "program1_2020_spring_class1", results[0].ProgramClasses[0].Classes[0].ClassId)
+	assert.EqualValues(t, "program1_2020_spring_class2", results[0].ProgramClasses[0].Classes[1].ClassId)
+	assert.EqualValues(t, "2020_summer", results[1].Semester.SemesterId)
+	assert.EqualValues(t, "program1", results[1].ProgramClasses[0].ProgramObj.ProgramId)
+	assert.EqualValues(t, "program1_2020_summer_final_review", results[1].ProgramClasses[0].Classes[0].ClassId)
+	assert.EqualValues(t, "program2", results[1].ProgramClasses[1].ProgramObj.ProgramId)
+}
+
+func GetOneSemesterTwoProgramsTwoClasses_Success(t *testing.T) {
```
ahsu1230, line 106: I think this needs to have Test infront of it in order to get tested.

constellations/orion/src/controllers/classes_by_semesters_controller_test.go
```diff
@@ -0,0 +1,223 @@
...
+		t.Errorf("unexpected error: %v\n", err)
+	}
+	assert.EqualValues(t, "2020_spring", results[0].Semester.SemesterId)
+	assert.EqualValues(t, "program1", results[0].ProgramClasses[0].ProgramObj.ProgramId)
+	assert.EqualValues(t, "program1_2020_spring_class1", results[0].ProgramClasses[0].Classes[0].ClassId)
+	assert.EqualValues(t, "program1_2020_spring_class2", results[0].ProgramClasses[0].Classes[1].ClassId)
+	assert.EqualValues(t, "program2", results[0].ProgramClasses[1].ProgramObj.ProgramId)
+}
+
+func GetTwoSemestersOneProgramOneClass(t *testing.T) {
```
ahsu1230, line 139: I think this needs to have Test infront of it in order to get tested.

constellations/orion/src/tests_integration/classes_by_semesters_integration_test.go
```diff
@@ -0,0 +1,34 @@
...
+	if err := json.Unmarshal(recorder.Body.Bytes(), &results); err != nil {
+		t.Errorf("unexpected error: %v\n", err)
+	}
+	assert.EqualValues(t, "2020_spring", results[0].Semester.SemesterId)
+	assert.EqualValues(t, "program1", results[0].ProgramClasses[0].ProgramObj.ProgramId)
+	assert.EqualValues(t, "program1_2020_spring_class1", results[0].ProgramClasses[0].Classes[0].ClassId)
+	assert.EqualValues(t, "program1_2020_spring_class2", results[0].ProgramClasses[0].Classes[1].ClassId)
+	assert.EqualValues(t, "2020_summer", results[1].Semester.SemesterId)
+
+	resetClassTables(t)
```
ahsu1230, line 33: very nice :)

constellations/orion/src/controllers/classes_by_semesters_controller.go
```diff
@@ -0,0 +1,118 @@
...
+
+	if err != nil {
+		c.Error(err)
+		c.String(http.StatusNotFound, err.Error())
+	} else {
+		c.JSON(http.StatusOK, &listResults)
+	}
+}
+
+func createProgramClassesForSemester(c *gin.Context, semesterId string, programs []domains.Program, classes []domains.Class, programMap map[string]domains.Program) []domains.ProgramClass {
```
ahsu1230, line 49: ok so looks like this is a helper function.

You can think of this helper function as only dealing with programs, semesters, classes, etc. And will return a list of `ProgramClasses`. Which is all great.

The function that calls THIS function `GetAllProgramsSemestersClasses` will deal with the JSON stuff. So you don't need to pass the gin.Context to this function. This function will ONLY deal with the domains. The caller will deal with JSON.

That way, it's neatly separated which function does what.

ahsu1230: I would have this function have TWO return types (unique in Go).

It will either return a list of ProgramClasses (if everything is ok) OR return an error (if something wrong happened)

constellations/orion/src/controllers/classes_by_semesters_controller.go
```diff
@@ -0,0 +1,118 @@
...
+	// Create map mapping programId to list of classes programClassMap
+	programClassMap := make(map[string][]domains.Class)
+
+	// For each class, get programId and put into programClassMap
+	for i := 0; i < len(classSlice); i++ {
+		programId := classSlice[i].ProgramId
+		if _, ok := programMap[programId]; !ok {
+			err := errors.New("programId not found in list of programs")
+			c.Error(err)
+			c.String(http.StatusInternalServerError, err.Error())
```
ahsu1230, line 68: Regarding the comment above with not passing in the context, instead, you can return an empty list and `err`. Then in the above function where you call this function. You can do something like this:

result, err := createProgramClassesForSemester(...)
if err != nil {
  // something bad happened!
  c.Error(err)
  c.String(...)
}

constellations/orion/src/controllers/classes_by_semesters_controller.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+	}
+
+	// Create list of ProgramClass
+	for i := 0; i < len(programs); i++ {
+		programId := programs[i].ProgramId
+		programClasses = append(programClasses, updateProgramClass(programs[i], programClassMap[programId]))
+	}
+
+	return programClasses
```
ahsu1230, line 93: And if everything is okay, you would `return programClasses, nil` so the error returned is nil.

constellations/orion/src/controllers/classes_by_semesters_controller_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+	"github.com/ahsu1230/mathnavigatorSite/constellations/orion/src/controllers/testUtils"
+	"github.com/ahsu1230/mathnavigatorSite/constellations/orion/src/domains"
+	"github.com/ahsu1230/mathnavigatorSite/constellations/orion/src/repos"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestOneSemesterOneProgramOneClass_Success(t *testing.T) {
+	// Mock 1 program, 1 semester, 1 class
+	testUtils.ProgramRepo.MockSelectAll = func(publishedOnly bool) ([]domains.Program, error) {
```
ahsu1230, line 17: This is how you define a function to return two things: a list of Programs, and an error. If no error (right below), it is nil.

constellations/orion/src/controllers/classes_by_semesters_controller.go
```diff
@@ -0,0 +1,125 @@
...
+
+	return programClasses, nil
+}
+
+func updateProgramClass(programObj domains.Program, classes []domains.Class) domains.ProgramClass {
+	return domains.ProgramClass{
+		ProgramObj: programObj,
+		Classes:    classes,
+	}
+}
```
ahsu1230, line 108: I don't think this function is too helpful. All it does is create an object. Yeah, it shrinks some lines, but if the function doesn't do enough, it's not worth refactoring into its own function.

constellations/orion/src/controllers/classes_by_semesters_controller.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	// Account for programs that are in map but have no classes
+	for key, _ := range programMap {
+		if _, ok := programClassMap[key]; ok {
+			continue
+		} else {
+			programClassMap[key] = []domains.Class{}
+		}
+	}
```
ahsu1230, line 92: This on the other hand, can be very worth to refactor it its own function.
Then this whole thing can just be `programClassMap := createProgramClassMap(...)`

constellations/orion/src/controllers/classes_by_semesters_controller_test.go
```diff
@@ -0,0 +1,223 @@
...
+	var results []domains.ProgramClassesBySemester
+	if err := json.Unmarshal(recorder.Body.Bytes(), &results); err != nil {
+		t.Errorf("unexpected error: %v\n", err)
+	}
+	assert.EqualValues(t, "2020_spring", results[0].Semester.SemesterId)
+	assert.EqualValues(t, "program1", results[0].ProgramClasses[0].ProgramObj.ProgramId)
+	assert.EqualValues(t, "program1_2020_spring_class1", results[0].ProgramClasses[0].Classes[0].ClassId)
+}
+
+func TestOneSemesterOneProgramOneClass_Failure(t *testing.T) {
```
ahsu1230, line 46: NoClasses?

constellations/orion/src/controllers/classes_by_semesters_controller_test.go
```diff
@@ -0,0 +1,223 @@
...
+	assert.EqualValues(t, "program1", results[0].ProgramClasses[0].ProgramObj.ProgramId)
+	assert.EqualValues(t, "program1_2020_spring_class1", results[0].ProgramClasses[0].Classes[0].ClassId)
+	assert.EqualValues(t, "program1_2020_spring_class2", results[0].ProgramClasses[0].Classes[1].ClassId)
+	assert.EqualValues(t, "2020_summer", results[1].Semester.SemesterId)
+	assert.EqualValues(t, "program1", results[1].ProgramClasses[0].ProgramObj.ProgramId)
+	assert.EqualValues(t, "program1_2020_summer_final_review", results[1].ProgramClasses[0].Classes[0].ClassId)
+	assert.EqualValues(t, "program2", results[1].ProgramClasses[1].ProgramObj.ProgramId)
+}
+
+func TestGetOneSemesterTwoProgramsTwoClasses_Success(t *testing.T) {
```
ahsu1230, line 106: OneClass?

ahsu1230: Mm better yet, maybe name this test names more specifically. Like this test is testing if a program doesn't have any classes. So maybe you can shrink to test JUST this scenario (i.e. 1 semester, 1 program, 0 classes)

constellations/orion/src/controllers/classes_by_semesters_controller_test.go
ahsu1230: Woops, which you already did here (see below). If there's something specific that you're testing here that's NOT in the other test, you should keep both. However, if there isn't anything different between these two tests, just keep one of them and reduce test redundancy.

constellations/orion/src/controllers/classes_by_semesters_controller_test.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+	assert.EqualValues(t, "2020_spring", results[0].Semester.SemesterId)
+	assert.EqualValues(t, "program1", results[0].ProgramClasses[0].ProgramObj.ProgramId)
+	assert.EqualValues(t, "program1_2020_spring_class1", results[0].ProgramClasses[0].Classes[0].ClassId)
+	assert.EqualValues(t, "program1_2020_spring_class2", results[0].ProgramClasses[0].Classes[1].ClassId)
+	assert.EqualValues(t, "program2", results[0].ProgramClasses[1].ProgramObj.ProgramId)
+}
+
+func GetTwoSemestersOneProgramOneClass(t *testing.T) {
+	// Mock 2 semesters, 1 program, 1 class, where one semester has no programs
```
ahsu1230, line 140: Hm I would just test 1 semester, but that semester has no program/classes. Simplifies this test because you only care about 1 semester.

constellations/orion/src/controllers/classes_by_semesters_controller_test.go
mythicat554: I think here I tested what happens if one of the repo functions returns an error (which would fail)

PR #195, opened 2020-07-26
Added school and graduation_year columns to users table

ChujiaGuo:


constellations/orion/src/repos/migrations/000030_add_school_at_users.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE users
+    ADD COLUMN school varchar(32);
```
ahsu1230, line 2: Hmm turns out "Winston churchill High school" is 29 characters long which is a bit too close for comfort.
Let's make this 64.

constellations/orion/src/tests_integration/user_integration_test.go
```diff
@@ -190,47 +190,55 @@ func createUser(id int) domains.User {
...
+			FirstName:      "Foo",
+			LastName:       "Bar",
+			MiddleName:     domains.NewNullString("Smith"),
+			Email:          "foobar@example.com",
+			Phone:          "555-555-0102",
+			IsGuardian:     false,
+			AccountId:      2,
+			Notes:          domains.NewNullString("notes3"),
+			School:         domains.NewNullString(""),
+			GraduationYear: domains.NewNullUint(0),
```
ahsu1230, line 228: ah nice

constellations/orion/src/domains/user.go
```diff
@@ -10,18 +10,20 @@ import (
...
+	FirstName      string       `json:"firstName" db:"first_name"`
+	LastName       string       `json:"lastName" db:"last_name"`
+	MiddleName     NullString   `json:"middleName" db:"middle_name"`
+	Email          string       `json:"email"`
+	Phone          string       `json:"phone"`
+	IsGuardian     bool         `json:"isGuardian" db:"is_guardian"`
+	AccountId      uint         `json:"accountId" db:"account_id"`
+	Notes          NullString   `json:"notes"`
+	School         NullString   `json:"school" db:"school"`
+	GraduationYear NullUint     `json:"graduationYear" db:"graduation_year"`
```
ahsu1230, line 26: Let's have some validation for graduation years. The graduation year (if not null) should be greater than 2000.
And School (if not null) should have only letters in them.

constellations/orion/src/domains/user.go
```diff
@@ -53,5 +57,14 @@ func (user *User) Validate() error {
 		return errors.New("invalid phone")
 	}
 
+	// School validation
+	if matches, _ := regexp.MatchString(`[^a-zA-Z\s]`, school); matches && user.School.Valid {
```
ahsu1230, line 61: the `[a-zA-Z]` gets used often in our codebase that we actually have a constant for it. Can you please replace this part with one of the patterns in domains/regex.go?

constellations/orion/src/domains/regex.go
```diff
@@ -7,6 +7,9 @@ const (
 	// Ensures at least one uppercase or lowercase letter
 	REGEX_LETTER = `[A-Za-z]`
 
+	// School names (alphabetic and some symbols)
```
ahsu1230, line 10: can you specify which symbols in the comments? for ease of use?

constellations/orion/src/domains/regex.go
```diff
@@ -7,6 +7,9 @@ const (
...
+	REGEX_SCHOOLS = `[^a-zA-Z\s.-]`
```
ahsu1230, line 11: hermm REGEX_LETTER_BASIC_SYMBOLS (i think).
There will most likely be another thing that will need this. So if there's ever a time we want to differentiate something school specific, we can rename to REGEX_SCHOOL. But for now, let's use something general.

constellations/orion/src/domains/user_valid_test.go
```diff
@@ -109,3 +109,70 @@ func TestValidPhone(t *testing.T) {
...
+	}
+	//Test valid school
+	user.School = domains.NewNullString("Churchill H.S.")
+	if err := user.Validate(); err != nil {
+		t.Errorf("Check was incorrect, got: %s, expected: nil", err.Error())
+	}
+	user.School = domains.NewNullString("Winston churchill High school")
+	if err := user.Validate(); err != nil {
+		t.Errorf("Check was incorrect, got: %s, expected: nil", err.Error())
+	}
```
ahsu1230, line 132: This is a redundant test to the Montgomery Blair one below so you can remove it (they test basically the same thing).

PR #201, opened 2020-07-30
adding column "ordering" to Semesters Table

austinhsu14:


constellations/orion/src/controllers/semester_controller_test.go
```diff
@@ -91,6 +95,8 @@ func TestGetSemester_Success(t *testing.T) {
 	}
 	assert.EqualValues(t, "2020_fall", semester.SemesterId)
 	assert.EqualValues(t, "Fall 2020", semester.Title)
+	assert.EqualValues(t, 1, semester.Ordering)
+
```
ahsu1230, line 99: nit: extra space

constellations/orion/src/repos/migrations/000032_add_ordering_at_semesters.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE semesters
+    ADD COLUMN ordering int unsigned;
```
ahsu1230, line 2: add DEFAULT 0 at the end. This column should not be NULL

constellations/orion/src/tests_integration/semester_integration_test.go
```diff
@@ -13,9 +13,9 @@ import (
 
 // Test: Create 3 Semesters and GetAll(false)
 func Test_CreateSemesters(t *testing.T) {
-	semester1 := createSemester("2020_spring", "Spring 2020")
-	semester2 := createSemester("2020_fall", "Fall 2020")
-	semester3 := createSemester("2020_winter", "Winter 2020")
+	semester1 := createSemester("2020_spring", "Spring 2020", 3)
+	semester2 := createSemester("2020_fall", "Fall 2020", 2)
+	semester3 := createSemester("2020_winter", "Winter 2020", 1)
```
ahsu1230, line 18: can you rename these to actually reflect what might happen in a school year.
i.e. 2020_spring comes first? then 2020_fall, then 2020_winter.
But we want to test if ordering works, so you'll have to rename one of them to 2019 or 2021 for example.

PR #202, opened 2020-07-31
First Time Importer script

ahsu1230:


ahsu1230:
Register endpoint in orion needs to be adjusted. Register class expects a student AND guardian information (from website). Here, we do not have guardian information.

PR #203, opened 2020-08-01
Input select

notewio:
@DanielLiu14 @ahsu1230


constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -220,23 +231,29 @@ export class ClassEditPage extends React.Component {
...
-                        onChange={(e) =>
+                        onChangeCallback={(e) =>
                             this.handleChange(e, "selectProgramId")
-                        }>
-                        {programOptions}
-                    </select>
+                        }
+                        required={true}
+                        options={programOptions}
+                        hasNoDefault={false}
```
ahsu1230, line 243: if it's false, you can just omit this field

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -220,23 +231,29 @@ export class ClassEditPage extends React.Component {
...
-                        onChange={(e) =>
+                        onChangeCallback={(e) =>
                             this.handleChange(e, "selectSemesterId")
-                        }>
-                        {semesterOptions}
-                    </select>
+                        }
+                        required={true}
+                        options={semesterOptions}
+                        hasNoDefault={false}
```
ahsu1230, line 255: same if false, can just omit

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -275,14 +292,17 @@ export class ClassEditPage extends React.Component {
...
-                        onChange={(e) =>
+                        onChangeCallback={(e) =>
                             this.handleChange(e, "selectLocationId")
-                        }>
-                        {locationOptions}
-                    </select>
+                        }
+                        required={true}
+                        options={locationOptions}
+                        hasNoDefault={false}
```
ahsu1230, line 304: can omit

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -300,14 +320,15 @@ export class ClassEditPage extends React.Component {
...
                         value={this.state.fullState}
-                        onChange={(e) => this.onChangeFullState(e)}>
-                        <option value="0">Normal</option>
-                        <option value="1">Almost Full</option>
-                        <option value="2">Full</option>
-                    </select>
+                        onChangeCallback={(e) => this.onChangeFullState(e)}
+                        required={true}
+                        options={fullStateOptions}
+                        hasNoDefault={false}
```
ahsu1230, line 330: can omit

constellations/gemini-admin/src/utils/inputSelect.js
```diff
@@ -0,0 +1,105 @@
+"use strict";
+require("./inputText.sass");
+import React from "react";
+import Checkbox from "../../assets/checkmark_green.svg";
+
+/*
+ * Props for the InputSelect Component:
```
ahsu1230, line 7: oh thanks for these docs

constellations/gemini-admin/src/utils/inputSelect.js
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * value: Current selected option
+ *
+ * onChangeCallback: Function that is called when the selection changes.
+ *
+ * required: Whether or not the selection  is required (omit if false).
+ *
+ * hasNoDefault: If true, an extra option -- Select an option -- will be added as the default.
+ *                  It cannot be reselected after another option is chosen.
+ *
```
ahsu1230, line 17: sorry can you add one more (optional) field.

If the list is empty, but it really shouldn't be. Something went wrong or a user action is required. For example, if I'm creating a class, but there's no program or semester to choose from. We should ask the user to first add a program/semester.

In this case, add another field `errorMessageIfEmpty` which takes in a string. If the list is empty, display this message. If the list is non-empty, do nothing.

constellations/gemini-admin/src/utils/inputSelect.js
```diff
@@ -0,0 +1,105 @@
...
+        var optionElements = options.map((option, index) => (
+            <option key={index} value={option.value}>
+                {option.displayName}
+            </option>
+        ));
+
+        var defaultOption = this.props.hasNoDefault ? (
+            <option disabled selected value>
+                {" "}
+                -- Select an option --{" "}
```
ahsu1230, line 67: Hm this is odd formatting. Anyway we can just get this to be: `{" "}-- Select an option --{" "}` or use HTML code for a single space? Like `&nbsp;`?

ahsu1230:
```
&nbsp;
-- Select an option --
&nbsp;
```

does that work? ^

constellations/gemini-admin/src/utils/inputText.sass
```diff
@@ -14,7 +14,7 @@
         display: flex
         align-items: center
 
-        input
+        input, select
```
ahsu1230, line 17: oh nice, resourceful

constellations/gemini-admin/src/utils/inputText.sass
```diff
@@ -14,7 +14,7 @@
         display: flex
```
ahsu1230, line 14: I think the h2 above gets its margins overriden in some pages.
Could you give the h2 a `margin: spacing2x 0`?

notewio:
> We'll also need this InputSelect in the following pages:
> * UserAFH
> * UserClass

How should I deal with the `required` field in these pages? should I just leave the (optional) in or add another field to hide it?

ahsu1230:
> > We'll also need this InputSelect in the following pages:
> > 
> > * UserAFH
> > * UserClass
> 
> How should I deal with the `required` field in these pages? should I just leave the (optional) in or add another field to hide it?

I think they the select Class / AFH should be required for both of those pages right? And make there be no default for both of them. So it should say `(required)` and once they select something from the list, the checkmark should appear.

You can move the "Enroll" button to below the InputSelect on both pages.

constellations/gemini-admin/src/user/userAFH.js
```diff
@@ -85,40 +86,42 @@ export class UserAFHPage extends React.Component {
...
-                        " " +
-                        afh.subject +
-                        " " +
-                        afh.timeString}
-                </option>
-            );
+            return {
+                value: afh.id,
+                displayName:
+                    moment(afh.date).format("1") +
```
ahsu1230, line 92: `I think you mean .format("l")` -> lowercase L for local time string

PR #205, opened 2020-08-02
intern polish

srenaxu:
@ahsu1230 @notewio

### DIFF
--- a/constellations/gemini-user/src/internship/internship.js
+++ b/constellations/gemini-user/src/internship/internship.js
@@ -24,7 +24,7 @@ export class InternshipPage extends React.Component {
                 </p>
                 <h1>Technology Stack</h1>
 
-                <div class="container-main">
+                <div className="container-main">
                     <img src={ReactImg} />
                     <p>
                         ReactJs, created by Facebook, is a popular web framework
@@ -33,7 +33,7 @@ export class InternshipPage extends React.Component {
                     </p>
                 </div>
 
-                <div class="container-main">
+                <div className="container-main">
                     <img src={GoLangImg} />
                     <p>
                         Golang, created by Google, is an open source programming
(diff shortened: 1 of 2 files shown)


srenaxu:
cathy's row, in the list of past interns, keeps getting messed up and i can't figure out why. at 1000px, RM and MIT were not aligned w the other schools but it was subtle

ahsu1230:
The changes you made aren't wrong. But I think I found a few things that need to change for this page anyway which will help you solve your edge case.

1) Change all your `class=...` to `className`

2) `<p>` stands for paragraph which is usually a block element (takes up a whole row). Since all these elements are next to each other, it's better to use an inline element like `<span>`. So inside each of intern rows, switch p => span.

3) If someone doesn't have a college, you can just leave it. You don't need an empty `<p>` or `<span>`.

4) To solve your formatting issue, I think this might be better:
```
    .intern-main
        display: flex
        flex-direction: row
        margin-bottom: $spacing3x

        & span
            &.name
                width: 128px
            &.school
                width: 360px
            &.college
                width: calc(100% - 128px - 360px)
```

What i'm doing above is giving the WHOLE row a margin bottom - that way each of the elements don't need their own margin bottom. Second, I'm defining the first two elements with a particular width and then the last one (college column) is everything except else. We can use math to calculate how large it should be and limit it to that width at all times. So the third column is a percentage based size and as long as the widths of name, school, and college add to 100%, it will never break formatting.

Now here as an example, I'm using hardcoded values like 128px and 360px. Please use variables instead so we have consistent spacing. To use variables with calc, refer to this: <https://stackoverflow.com/questions/17982111/sass-variable-in-css-calc-function>

ahsu1230:
Looks good to merge!

PR #206, opened 2020-08-02
Updating filler.go in orionFakeDataFiller

ChujiaGuo:


ahsu1230:
Sweet! So I added a few things:

1) Fixed the sessions part - it's a very nuanced bug. It's the only endpoint that takes in a list instead of just one at a time... because in the webclient we usually add many sessions at a time. I forgot about this feature lol.

2) You're having issues with filling users and accounts, etc. because you must first add an account! Then add a user to that account because `user` relies on `accountId`! And so on.

3) I've added a reset db script in the `resetOrionDb` folder in `clients`. This should be super useful for you since you use a running MySQL server. For others using Docker, this isn't as needed because you can simply decompose your docker stuff, and rebuild it with a few commands.

Take a look at how to use the reset script! It should be easy to then check if your filler is working (not too many errors everywhere).

constellations/clients/orionFakeDataFiller/filler.go
```diff
@@ -228,28 +238,87 @@ func runFiller(hostAddress string) {
...
+		"Smith",
+		"",
+		"emailaddress1@gmail.com",
+		"301-123-4567",
+		0,
+		1,
+		"notes1",
+		"schoolone",
+		2001,
+	)
```
ahsu1230, line 279: can you make 2 users for account1. Account2 having only one user is fine too. Try to simulate what it will look like in "real life".

So account1 is for one family, while account2 is for another family.

ahsu1230: Account1 could have a parent and 2 kids or something? And if they're a guardian, they shouldn't have a school / grad year

constellations/clients/orionFakeDataFiller/filler.go
```diff
@@ -228,28 +238,87 @@ func runFiller(hostAddress string) {
...
+		"notes1",
+		1,
+	)
+
+	createTransaction(
+		hostAddress,
+		101,
+		"pay_cash",
+		"notes2",
+		2,
```
ahsu1230, line 309: add one more transaction that's a charge (don't mind which one)

ahsu1230: it would be great if one of these accounts had a negative a balance (i.e. stills owe us)

constellations/clients/orionFakeDataFiller/filler.go
```diff
@@ -318,25 +387,80 @@ func createClass(hostAddress, programId, semesterId, classKey, classId, location
...
+}
+
+func createUser(hostAddress, first_name, last_name, middle_name, email, phone string, is_guardian int, account_id int, notes, school string, graduation_year int) error {
+	userBody := strings.NewReader(fmt.Sprintf(`{
+		"firstName": "%s",
+		"lastName": "%s",
+		"middleName": "%s",
+		"email": "%s",
+		"phone": "%s",
+		"isGuardian": %d,
```
ahsu1230, line 434: shouldn't isGuardian be a boolean?

constellations/clients/orionFakeDataFiller/filler.go
```diff
@@ -228,28 +238,123 @@ func runFiller(hostAddress string) {
...
-		"id_2",
+		"ap_calculus_2020_fall_class1",
 		"true",
+		2,
+	)
+
+	// Create Accounts
+	createAccount(
+		hostAddress,
+		"emailaddress1@gmail.com",
```
ahsu1230, line 258: i think to make it more clear who is the owner of this account, make this email more specific. i.e. joesmith@gmail.com. And the user's email would need to change as well

constellations/clients/orionFakeDataFiller/filler.go
```diff
@@ -228,28 +238,123 @@ func runFiller(hostAddress string) {
...
+		"jhdgjhddjhdjuj",
+	)
+	createAccount(hostAddress,
+		"emailaddress2@gmail.com",
```
ahsu1230, line 262: make this email easily reflect who owns this account (for our readability sake)

constellations/clients/orionFakeDataFiller/filler.go
```diff
@@ -228,28 +238,123 @@ func runFiller(hostAddress string) {
...
+		"2redssssa",
+	)
+
+	// Create Users
+	createUser(
+		hostAddress,
+		"Joe",
+		"Smith",
+		"",
+		"emailaddress1@gmail.com",
```
ahsu1230, line 272: joesmith@gmail.com

constellations/clients/orionFakeDataFiller/filler.go
```diff
@@ -228,28 +238,123 @@ func runFiller(hostAddress string) {
...
+		"schoolone",
+		2001,
+	)
+
+	createUser(
+		hostAddress,
+		"Joe",
+		"Smith",
+		"Mom",
+		"emailaddress1@gmail.com",
```
ahsu1230, line 286: marysmith-mom@gmail.com

constellations/clients/orionFakeDataFiller/filler.go
```diff
@@ -228,28 +238,123 @@ func runFiller(hostAddress string) {
...
+	createUser(
+		hostAddress,
+		"Billy",
+		"Bob",
+		"Joe Dad",
+		"emailaddress2@gmail.com",
+		"301-123-4568",
+		true,
+		2,
+		"notes2",
```
ahsu1230, line 318: you don't always have to include notes. So this can be "" if you want

constellations/clients/orionFakeDataFiller/filler.go
```diff
@@ -228,28 +238,123 @@ func runFiller(hostAddress string) {
...
+		hostAddress,
+		"Billy",
+		"Bob",
+		"Joe",
+		"emailaddress2@gmail.com",
+		"301-123-4568",
+		false,
+		2,
+		"notes2",
+		"schooltwo",
```
ahsu1230, line 305: I think this will also be useful if it was a "real" high school name.
`Montgomery Richard HS` or something so when we display on the UI (front-end website), it looks closer to what we expect. `schooltwo` seems like it's an id or something

constellations/clients/orionFakeDataFiller/filler.go
```diff
@@ -228,28 +238,123 @@ func runFiller(hostAddress string) {
...
+		2002,
+	)
+
+	createUser(
+		hostAddress,
+		"Billy",
+		"Bob",
+		"Joe Dad",
```
ahsu1230, line 313: Dad

constellations/clients/orionFakeDataFiller/filler.go
```diff
@@ -228,28 +238,123 @@ func runFiller(hostAddress string) {
...
+		"emailaddress2@gmail.com",
+		"301-123-4568",
```
ahsu1230, line 315: could you also make the phone numbers different? Maybe down the line we want to search for a user's phone number.

PR #207, opened 2020-08-02
home dashboard

srenaxu:
@ahsu1230 @notewio

### DIFF
--- a/constellations/gemini-admin/src/home/home.js
+++ b/constellations/gemini-admin/src/home/home.js
@@ -4,9 +4,23 @@ import React from "react";
 import API from "../api.js";
 import { Link } from "react-router-dom";
 
+const sectionDisplayNames = {
+    class: "Unpublished Classes",
+    registration: "New Registrations",
+    user: "New Users",
+    unpaid: "Unpaid Accounts",
+};
+
 export class HomePage extends React.Component {
     state = {
         classes: [],
+        currentSection: "class",
+    };
+
+    changeSection = (sectionName) => {
+        this.setState({
+            currentSection: sectionName,
+        });
     };
 
     componentDidMount() {
@@ -31,11 +45,36 @@ export class HomePage extends React.Component {
             <div id="view-home">
                 <h1>Administrator Dashboard</h1>
 
+                <div className="tabs">
+                    <TabButton
+                        onChangeTab={this.changeSection}
+                        highlight={this.state.currentSection == "class"}
+                        section={"class"}
+                    />
+                    <TabButton
+                        onChangeTab={this.changeSection}
+                        highlight={this.state.currentSection == "registration"}
+                        section={"registration"}
+                    />
+                    <TabButton
+                        onChangeTab={this.changeSection}
+                        highlight={this.state.currentSection == "user"}
+                        section={"user"}
+                    />
+                    <TabButton
+                        onChangeTab={this.changeSection}
+                        highlight={this.state.currentSection == "unpaid"}
+                        section={"unpaid"}
+                    />
+                </div>
+
                 <div className="section">
                     <div className="container-class">
                         <h3 className="section-header">Unpublished Classes</h3>{" "}
                         <button id="publish">
-                            <Link to={"/classes"}>View Details to Publish</Link>
+                            <Link to={"/classes"}>
+                                View All Classes to Publish
+                            </Link>
                         </button>
                     </div>
 
@@ -47,62 +86,28 @@ export class HomePage extends React.Component {
 
                 <div className="section">
                     <h3 className="section-header">New Users</h3>
-                    <div className="container-user">
-                        <p>
-                            Lorem ipsum dolor sit amet, vim ei tota dicant
-                            interpretaris, sea consulatu scripserit ei, pri eu
-                            accumsan contentiones. Usu ei iriure deleniti. Cum
-                            ut periculis laboramus referrentur. Vis ea laoreet
-                            imperdiet deterruisset, probo pertinax iudicabit qui
-                            ea. Duo liber quodsi contentiones ne, vel ei movet
-                            conceptam, id his quod iriure feugiat. Qui ex vide
-                            labitur volumus. His cu fugit adolescens
-                            voluptatibus, et per torquatos interesset. Iusto
-                            deleniti invenire id cum. Te saepe alterum
-                            appellantur pro, in est errem dicant suscipit,
-                            oblique argumentum sed et. Quo cu partem inermis,
-                            vix et minimum vivendum, ut illud delectus eos. Mei
-                            te alia justo, per amet nemore quodsi cu, mel
-                            eruditi copiosae contentiones cu. Sed ea quando
-                            mediocrem, soleat dolorum no ius. Eu electram
-                            iracundia mnesarchum his, eum ut hinc latine. Est
-                            simul definiebas ut, debitis invenire eu usu. Eu
-                            eros appetere mel, nullam delenit tincidunt duo at.
-                            Eos option appetere torquatos in, cu debitis
-                            singulis principes usu.
-                        </p>
-                    </div>
                 </div>
 
                 <div className="section">
                     <h3 className="section-header">New Registrations</h3>
-                    <div className="container-registration">
-                        <p>
-                            Lorem ipsum dolor sit amet, vim ei tota dicant
-                            interpretaris, sea consulatu scripserit ei, pri eu
-                            accumsan contentiones. Usu ei iriure deleniti. Cum
-                            ut periculis laboramus referrentur. Vis ea laoreet
-                            imperdiet deterruisset, probo pertinax iudicabit qui
-                            ea. Duo liber quodsi contentiones ne, vel ei movet
-                            conceptam, id his quod iriure feugiat. Qui ex vide
-                            labitur volumus. His cu fugit adolescens
-                            voluptatibus, et per torquatos interesset. Iusto
-                            deleniti invenire id cum. Te saepe alterum
-                            appellantur pro, in est errem dicant suscipit,
-                            oblique argumentum sed et. Quo cu partem inermis,
-                            vix et minimum vivendum, ut illud delectus eos. Mei
-                            te alia justo, per amet nemore quodsi cu, mel
-                            eruditi copiosae contentiones cu. Sed ea quando
-                            mediocrem, soleat dolorum no ius. Eu electram
-                            iracundia mnesarchum his, eum ut hinc latine. Est
-                            simul definiebas ut, debitis invenire eu usu. Eu
-                            eros appetere mel, nullam delenit tincidunt duo at.
-                            Eos option appetere torquatos in, cu debitis
-                            singulis principes usu.
-                        </p>
-                    </div>
                 </div>
             </div>
+        );
+    }
+}
+
+class TabButton extends React.Component {
+    render() {
+        let highlight = this.props.highlight;
+        let section = this.props.section;
+        let displayName = sectionDisplayNames[section];
+
+        return (
+            <button
+                className={highlight ? "active" : ""}
+                onClick={() => this.props.onChangeTab(section)}>
+                {displayName}
+            </button>
         );
     }
 }
--- a/constellations/gemini-admin/src/home/home.sass
+++ b/constellations/gemini-admin/src/home/home.sass
@@ -10,6 +10,30 @@
         margin-top: $spacing10x
         color: $colorTurquoise
 
+    .tabs 
+        display: flex
+        flex-direction: row
+        align-items: center
+        border-bottom: 1px solid $colorGrayLight
+
+        .active
+            border-bottom: 2px solid $colorTurquoise
+
+    .tabs button 
+        width: 33.333%
+        height: 100%
+        padding: $spacing2x
+        background-color: $colorWhite
+        border-bottom: 2px solid transparent
+        font-size: $fontSize1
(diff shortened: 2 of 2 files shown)


constellations/gemini-admin/src/home/home.js
```diff
@@ -1,108 +1,110 @@
...
+    };
+
+    render() {
+        let highlight = this.props.highlight;
+        let section = this.props.section;
+        let displayName = sectionDisplayNames[section];
+        let numNotif = this.props.buttonNum;
+
+        if (numNotif == 0) {
+            this.setState({
```
srenaxu, line 90: having trouble here. state variable isZero isn't updating so "zero" can't be added to the className for displaNotif. need to add "zero" for css

constellations/gemini-admin/src/home/home.js
```diff
@@ -1,108 +1,110 @@
...
+                        onChangeTab={this.changeSection}
+                        highlight={this.state.currentSection == "class"}
+                        section={"class"}
+                        buttonNum={<ClassesNotif />}
+                    />
+                    <TabButton
+                        onChangeTab={this.changeSection}
+                        highlight={this.state.currentSection == "user"}
+                        section={"user"}
+                        //buttonNum = {<UsersNotif />}cnewUsers.length in homeUsers
```
ahsu1230, line 56: I think this is correct. The reason buttonNum is always 0 is because you need it to be a certain value.
`buttonNum` from what it looks like is supposed to be an integer.

So you can simply do something like this: `buttonNum = {10}` or `buttonNum={newUsers.length}`

constellations/gemini-admin/src/home/homeClasses.js
```diff
@@ -0,0 +1,83 @@
...
+                    <div className="list-header">Class ID</div>
+                    {publishedMessage}
+                    <ul>{unpublishedClasses}</ul>
+                </div>
+            </div>
+        );
+    }
+}
+
+export class ClassesNotif extends React.Component {
```
ahsu1230, line 62: what is ClassesNotif supposed to represent?

constellations/gemini-admin/src/home/home.js
```diff
@@ -2,107 +2,126 @@
...
+    };
+
+    componentDidMount = () => {
+        if (this.props.buttonNum == 0) {
+            this.setState({
+                isZero: true,
+            });
+        }
+        console.log("isZero state is " + this.state.isZero);
+    };
```
notewio, line 105: i think it's not working because `componentDidMount` only runs once, and because `buttonNum` comes from an async function this all runs before the actual number is passed in. I'm not sure if using state is really necessary, having isZero as a local variable in render might be easier

tip: `setState` is an async function which is why the `console.log` is always showing false. you can things run after setState by passing in a function as the second argument, e.g.
```
this.setState({
    isZero: true,
}, () => {
    console.log("isZero state is " + this.state.isZero)
})
```

ahsu1230:
The newUsers and newUserRegistrations for classes/afh endpoints are in! So you can try them out to see how this all fits!
The endpoints are:

```
api/users/new
api/user-classes/new
api/userafhs/new
```

constellations/gemini-admin/src/home/home.js
ahsu1230: +1 I would get rid of this.state.isZero and instead, you can just always check if this.props.buttonNum is equal to 0. By introducing a new state variable that is similar to this props variable, you introduce weird states you could potentially be in (i.e. `buttonNum` is 0, but there's a small window of time that `isZero` is not 0).

So below, just have:
```
const numNotif = this.props.buttonNum;
...
<div className={"notif" + (numNotif == 0 ? " zero" : "")}>
                {numNotif}
</div>
```

constellations/gemini-admin/src/home/home.js
```diff
@@ -2,107 +2,151 @@
...
+            });
+        });
+
+        API.get("api/accounts/unpaid").then((res) => {
+            const accounts = res.data;
+            this.setState({
+                unpaidAccounts: accounts,
             });
         });
     };
```
tonywu315, line 71: Instead of multiple separate API calls, you could use `axios` to combine them into one. This way, you can have only one catch block. It would look something like this:
```
const apiCalls = [
    API.get("api/unpublished"),
    API.get("api/users/new"),
    // etc
];

axios
    .all(apiCalls)
    .then(
        axios.spread((...responses) => {
            this.setState({
                unpublishedClasses: responses[0].data.classes,
                newUsers: responses[1].data,
                // etc
            })
        })
    )
    .catch((err) => {
        console.log("Error: api call failed. " + err.message);
    });
```

constellations/gemini-admin/src/home/home.js
```diff
@@ -2,107 +2,151 @@
...
-
-                    <div className="class-section">
-                        <div className="list-header">Class ID</div>
-                        <ul>{unpublishedClasses}</ul>
-                    </div>
+                <div className="tabs">
+                    <TabButton
+                        onChangeTab={this.changeSection}
+                        highlight={this.state.currentSection == "class"}
+                        section={"class"}
```
tonywu315, line 94: You can use `"class"` instead of `{"class"}`.

constellations/gemini-admin/src/home/home.js
```diff
@@ -2,107 +2,151 @@
...
-                </div>
+                <div className="showSection">{sectionComponent}</div>
             </div>
         );
     }
 }
+
+class TabButton extends React.Component {
+    render() {
+        let highlight = this.props.highlight;
```
tonywu315, line 128: Since these don't change, `const` is better.

constellations/gemini-admin/src/home/home.js
```diff
@@ -2,107 +2,151 @@
...
+        let displayName = sectionDisplayNames[section];
+        let numNotif = this.props.buttonNum;
+
+        var isZero = "";
+
+        if (numNotif == 0) {
+            isZero = "notif zero";
+        } else {
+            isZero = "notif";
+        }
```
tonywu315, line 139: This can be simplified using a ternary expression:
`const isZero = numNotif == 0 ? "notif zero" : "notif"`.

constellations/gemini-admin/src/home/homeAccounts.js
```diff
@@ -0,0 +1,80 @@
...
+
+    render() {
+        let unpaidAcc = this.state.unpaidAccounts.map((row, index) => {
+            return (
+                <li className="container-flex" key={index}>
+                    <AccountInfo accountId={row.id} />
+                    <div className="other">{row.id} </div>
+                </li>
+            );
+        });
```
tonywu315, line 30: `row` has the fields `account` and `balance`, so `row.id` is undefined.

constellations/gemini-admin/src/home/homeRegistrations.js
```diff
@@ -0,0 +1,128 @@
...
+import { Link } from "react-router-dom";
+import { getFullName } from "../utils/userUtils.js";
+
+export class HomeTabSectionRegistrations extends React.Component {
+    state = {
+        classReg: [],
+        afhReg: [],
+    };
+
+    componentDidMount() {
```
tonywu315, line 14: You should use `axios` here too.

constellations/gemini-admin/src/home/home.js
```diff
@@ -2,107 +2,169 @@
...
+    class: "Unpublished Classes",
+    user: "New Users",
+    registration: "New Registrations",
+    unpaid: "Unpaid Accounts",
+};
 
 export class HomePage extends React.Component {
     state = {
-        classes: [],
+        currentSection: "class",
```
ahsu1230, line 19: when we use constants, it's best to define constant variables at the top of the file.

So at the top I would have:

const TAB_CLASSES = "classes";
const TAB_REGISTRATIONS = "registrations";
const TAB_USERS = "users";
const TAB_UNPAID = "unpaid";
then here use: selectedTab: TAB_CLASSES

It's better to do things this way because you force tab selection to use "typed" variables instead of a value where you are prone to typos.
For example, if you had currentSection: "clases", you might not notice this problem until your tab selection doesn't work. However, if you did selectedTab: TAB_CLASES you would get a compile error and immediately alert you!

constellations/gemini-admin/src/home/home.js
```diff
@@ -2,107 +2,169 @@
...
+
+        if (this.state.currentSection == "class") {
+            sectionComponent = <HomeTabSectionClasses />;
+        } else if (this.state.currentSection == "user") {
+            sectionComponent = <HomeTabSectionUsers />;
+        } else if (this.state.currentSection == "registration") {
+            sectionComponent = <HomeTabSectionRegistrations />;
+        } else {
+            sectionComponent = <HomeTabSectionAccounts />;
+        } //unpaid account
```
ahsu1230, line 84: can remove comment

constellations/gemini-admin/src/home/home.js
```diff
@@ -2,107 +2,169 @@
...
         });
     };
 
     render() {
-        let unpublishedClasses = this.state.classes.map((row, index) => {
-            return <li key={index}> {row.classId} </li>;
-        });
+        var sectionComponent = <div></div>;
+
+        if (this.state.currentSection == "class") {
```
ahsu1230, line 76: this.state.currentSection == TAB_CLASSES

constellations/gemini-admin/src/home/home.js
```diff
@@ -2,107 +2,169 @@
...
-                    </div>
-
-                    <div className="class-section">
-                        <div className="list-header">Class ID</div>
-                        <ul>{unpublishedClasses}</ul>
-                    </div>
+                <div className="tabs">
+                    <TabButton
+                        onChangeTab={this.changeSection}
+                        highlight={this.state.currentSection == "class"}
```
ahsu1230, line 93: this.state.currentSection == TAB_CLASSES

constellations/gemini-admin/src/home/home.js
```diff
@@ -2,107 +2,169 @@
...
+
+class TabButton extends React.Component {
+    render() {
+        const highlight = this.props.highlight;
+        const section = this.props.section;
+        const displayName = sectionDisplayNames[section];
+        const numNotif = this.props.buttonNum;
+
+        const isZero = numNotif == 0 ? "notif zero" : "notif";
+        console.log("isZero state is " + isZero);
```
ahsu1230, line 134: can get rid of print statement

constellations/gemini-admin/src/home/home.js
```diff
@@ -2,107 +2,169 @@
...
+                publishMessage = (
+                    <p>All classes have been successfully published!</p>
+                );
+            } else if (emptySection == "user") {
+                publishMessage = <p>There are no new users!</p>;
+            } else if (emptySection == "registration") {
+                publishMessage = <p>There are no registrations!</p>;
+            } else {
+                publishMessage = <p>All accounts have paid!</p>;
+            } //unpaid account
```
ahsu1230, line 165: consider a switch statement. Might be a bit cleaner code.

constellations/gemini-admin/src/home/home.js
ahsu1230: also consider using a switch statement here, might be cleaner code

constellations/gemini-admin/src/home/homeAccounts.js
```diff
@@ -0,0 +1,105 @@
...
+                <div className="container-class">
+                    <h3 className="section-header">Unpaid Accounts</h3>{" "}
+                    <button className="view-details">
+                        <Link to={"/accounts"}>View All Accounts</Link>
+                    </button>
+                </div>
+
+                <div className="class-section">
+                    <div className="container-flex">
+                        <div className={"list-header" + " name"}>Account</div>
```
ahsu1230, line 60: You can just do `lassName={"list-header name"}`

constellations/gemini-admin/src/home/homeAccounts.js
```diff
@@ -0,0 +1,105 @@
...
+                    <div className="name">
+                        <p id="account"> Account Id: {row.account.id} </p>
+                        <AccountInfo accountId={row.account.id} />
+                    </div>
+                    <div className="email">{row.account.primaryEmail} </div>
+                    <div className="balance">
+                        {"-$"}
+                        {Math.abs(row.balance)}
+                    </div>
+                    <div className="fromNow">
```
ahsu1230, line 37: class names are not case sensitive. I think you should use `from-now` instead.

constellations/gemini-admin/src/home/homeClasses.js
```diff
@@ -0,0 +1,69 @@
...
+                    <button className="view-details">
+                        <Link to={"/classes"}>
+                            View All Classes to Publish{" "}
+                        </Link>
+                    </button>
+                </div>
+
+                <div className="class-section">
+                    <div className="container-flex">
+                        <div className={"list-header" + " width50"}>
```
ahsu1230, line 53: "list-header width50"

constellations/gemini-admin/src/home/homeUsers.js
```diff
@@ -0,0 +1,71 @@
...
+                        <Link to={"/users/" + row.id + "/edit"}>
+                            {"View >"}
+                        </Link>
+                    </div>
+                </li>
+            );
+        });
+
+        return (
+            <div className="sectionDetails">
```
ahsu1230, line 44: section-details. Classes are not case sensitive.

PR #208, opened 2020-08-02
delete columns

mythicat554:


constellations/orion/src/controllers/publish_controller.go
```diff
@@ -19,38 +19,18 @@ func GetAllUnpublished(c *gin.Context) {
...
 
 	unpublishedDomains := domains.UnpublishedDomains{
-		Programs:  programList,
-		Classes:   classList,
-		Locations: locationList,
-		Achieves:  achieveList,
-		Semesters: semesterList,
-		Sessions:  sessionList,
+		Classes:  classList,
+		Achieves: achieveList,
```
ahsu1230, line 33: should achieveList also get removed? Looks like achievements also have a PublishedAt column

ahsu1230:
This can be merged if the previous msg wasn't clear lol

PR #210, opened 2020-08-05
Add programs and class pages

tonywu315:


constellations/gemini-user/src/class/class.js
```diff
@@ -0,0 +1,219 @@
...
+        classObj: {},
+        sessions: [],
+        program: {},
+        semester: {},
+        location: {},
+    };
+
+    componentDidMount = () => {
+        const classId = this.props.classId;
+        const apiClassCalls = [
```
ahsu1230, line 20: apiCallsPerClass

constellations/gemini-user/src/class/class.js
```diff
@@ -0,0 +1,219 @@
...
+                                    err.message
+                            )
+                        );
+                })
+            )
+            .catch(
+                (err) =>
+                    console.log("Error: could not fetch class: " + err.message)
+                // TODO: error page?
+            );
```
ahsu1230, line 68: This looks a bit crazy (especially with the indents), anyway we can modularize these into separate functions? 

Maybe something like:

```
fetchClassAndSessions(apiClassCalls)
    .then( axios.spread((...responses) => {
        fetchOtherData(responses)
            .then(...)
            .catch(...);
    })
    .catch((err) => ...);

fetchOtherData = (responses) => {
   ... get the programs, semesters, and locations ...
}
```

^ something like above. You don't have to do it completely that way. I'm just trying to provide an alternative with how you can organize better. With above, it's clear what the API calling order is. First we get Classes&Sessions, then we get other random stuff based on those responses.
Also, it's decently clear what will happen when failures occur. You can see which `catch` is for which set of API calls.

constellations/gemini-user/src/modals/modal.js
```diff
@@ -0,0 +1,49 @@
+"use strict";
+require("./modal.sass");
+import React from "react";
+import srcClose from "../../assets/close_black.svg";
+
+const classnames = require("classnames");
+
+export class Modal extends React.Component {
```
ahsu1230, line 8: do you mind adding docs to this Modal?
Like what props values are available and what they do. And then a simple example of how to use it.

Thanks!!

ahsu1230: props to list:
content, modalClassName, show, persistent, withClose, onDismiss,

constellations/gemini-user/src/programs/programCard.sass
```diff
@@ -0,0 +1,45 @@
...
+        width: $propUnit6x
+        margin: 0
+        color: $colorWhite
+        line-height: $lineHeight1
+        font-size: $fontSize1
+
+    h3
+        margin: $spacing1x 0 $spacing4x
+        color: $colorWhite
+        line-height: $fontSize0
```
ahsu1230, line 27: line-height -> font-size -> color

constellations/gemini-user/src/programs/programCard.sass
```diff
@@ -0,0 +1,45 @@
...
+    @media screen and (max-width: $mediaWidth1)
+        display: block
+        margin: 0 auto $cardSpacing
+    
+    h2
+        width: $propUnit6x
+        margin: 0
+        color: $colorWhite
+        line-height: $lineHeight1
+        font-size: $fontSize1
```
ahsu1230, line 22: width, margin, line-height, font-size, color

constellations/gemini-user/src/programs/programModal.js
```diff
@@ -0,0 +1,48 @@
...
+        return times
+            .split(",")
+            .map((time, index) => <span key={index}>{time.trim()}</span>);
+    };
+
+    render = () => {
+        const classObj = this.props.classObj;
+        const name =
+            classObj.classKey +
+            ["", " (ALMOST FULL)", " (FULL)"][classObj.fullState];
```
ahsu1230, line 35: Can you move this list to a `constants.js` file in the utils folder? I think you might need to also include this information in the ClassPage. Maybe place it right above the "Grades" part? And if it's full or almost-full, make the font red.

And maybe name it like this: `FULL_STATE_DISPLAY_NAMES`. For "global" constants, we like to use upper-capitalized snake_case.

```
FULL_STATE_DISPLAY_NAMES = ["", " (ALMOST FULL)", " (FULL)"]
```

constellations/gemini-user/src/programs/programs.sass
```diff
@@ -0,0 +1,81 @@
...
+    .section
+        margin-bottom: $cardSpacing
+
+        .section-title
+            margin-bottom: $cardSpacing
+            border-bottom: 1px solid $colorGrayLight
+            color: $colorPrimary
+            font-family: $fontFamilyRockwell
+            line-height: 88px
+            font-size: $fontSize2
```
ahsu1230, line 78: I usually like to have size stuff first then color

line-height -> font-family -> font-size -> color

tonywu315: In the css_ordering file, color goes before text.

constellations/gemini-user/src/class/classSchedule.js
```diff
@@ -0,0 +1,46 @@
...
+            const times =
+                startTime.format("dddd h:mma") +
+                " - " +
+                endTime.format("h:mma");
+
+            const canceled = session.canceled;
+            var state = "row";
+            if (moment().isAfter(endTime))
+                state += canceled ? " pink" : " gray";
+            else state += canceled ? " red" : "";
```
ahsu1230, line 20: what do the different states here mean?

constellations/gemini-user/src/class/classSchedule.sass
```diff
@@ -0,0 +1,45 @@
...
+            width: 100%
+        
+    .red span
+        color: red
+    
+    .pink span
+        color: pink
+
+    .gray span
+        color: $colorGrayLight
```
ahsu1230, line 22: I'd prefer if the class names told us about the state
like red means "cancelled", gray means "cancelled but X" and pink means "cancelled but Y"

Also could you also use the same red as we use in the admin site? I forgot what the hex is.

constellations/gemini-user/src/programs/programModal.sass
```diff
@@ -0,0 +1,85 @@
...
+        margin-right: $spacing2x
+        color: $colorPrimary
+
+        @media screen and (max-width: $mediaWidth1)
+            display: block
+            width: auto
+            margin: 0
+            text-align: center
+    
+    .red
```
ahsu1230, line 66: same here

constellations/gemini-user/src/programs/programModal.sass
```diff
@@ -0,0 +1,85 @@
...
+        display: inline-flex
+        flex-flow: column nowrap
+        width: $propUnit7x
+    
+        @media screen and (max-width: $mediaWidth1)
+            display: flex
+            text-align: center
+
+    a
+        float: right
```
ahsu1230, line 79: with the introduction of display-flex, we should need `float` anymore

ahsu1230: well more like it's preferable NOT to use it anymore because display: flex is more reliable.

ahsu1230: clarification. It's okay to use if the element is the ONLY child in an element.
However, in this case, there are siblings to this `<a>`, so we should use a flexbox here.

constellations/gemini-user/src/programs/programs.js
```diff
@@ -0,0 +1,90 @@
...
+
+export class ProgramsPage extends React.Component {
+    state = {
+        semesters: [],
+        programClassesMap: {},
+    };
+
+    componentDidMount = () => {
+        // TODO: Change to below line when published_at is removed from all tables except classes
+        // API.get("api/classesbysemesters?published=true").then((res) => {
```
ahsu1230, line 16: I believe the published part of the endpoint is ready. The published_at stuff have been removed

PR #213, opened 2020-08-08
User account page

notewio:


constellations/gemini-user/src/account/account.js
```diff
@@ -0,0 +1,658 @@
...
+        }
+
+        return (
+            <div id="view-account">
+                <h1>Your Account</h1>
+
+                <div id="tab-container">{tabButtons}</div>
+
+                {tabContent}
+            </div>
```
ahsu1230, line 232: I like this though, it's super tight!

constellations/gemini-user/src/account/account.js
```diff
@@ -0,0 +1,658 @@
...
+                        toggleTabCallback={this.toggleViewAllClasses}
+                    />
+                );
+                break;
+            case "payment":
+                tabContent = (
+                    <PaymentTab transactions={this.state.transactions} />
+                );
+                break;
+        }
```
ahsu1230, line 223: Ok sweet, this is pretty well done actually. My only gripe is this one javascript file is super large for what it can be.
I think it would be better if we had a couple of files instead of one large file. And that way, each component can keep better track of their own isolated content.

So have a `src/account/settings.js`, `src/account/registrations.js`, `src/account/payments.js` files containing the SettingsTab, RegistrationXTab, and PaymentTab respectively. And this file `account.js` will import components from those 3 files. Now each file is only 150-250 lines and much easier to see what's being managed by what.

For example, transactions are only used in the PaymentTab. So technically, PaymentTab can be its own component and its componentDidMount() function can fetch the transactions. So now AccountPage doesn't have to worry about transactions, it just has to manage when to show PaymentTab.

So now all transactions stuff can be removed from AccountPage and placed into PaymentTab. In order to get all transactions for an account, PaymentTab needs an `accountId`, which `AccountPage` can simply just pass as a parameter to PaymentTab.

```
let tabContent = <PaymentTab acccountId={this.state.accountId}/>
```

Inside PaymentTab, we fetch the transactions, and display what we need to display. The same can be said for the other Tab components. They all have information only they care about, so let those components handle those by themselves. Have AccountPage only manage which tab to show to the user.

constellations/gemini-user/src/account/account.js
```diff
@@ -0,0 +1,95 @@
...
+    RegistrationsTab,
+    RegistrationsTabMain,
+    RegistrationsTabAllClasses,
+} from "./registrations.js";
+import { PaymentTab } from "./payment.js";
+
+export class AccountPage extends React.Component {
+    state = {
+        id: 0,
+        selectedTab: "settings",
```
ahsu1230, line 18: when we use constants, it's best to define constant variables at the top of the file.

So at the top I would have:
```
const TAB_SETTINGS = "settings";
const TAB_REGISTRATIONS = "registrations";
const TAB_PAYMENT = "payment";
```

then here use: `selectedTab: TAB_SETTINGS`

It's better to do things this way because you force tab selection to use "typed" variables instead of a value where you are prone to typos.
For example, if you had `selectedTab: "setings"`, you might not notice this problem until your tab selection doesn't work. However, if you did `selectedTab: TAB_SETINGS` you would get a compile error and immediately alert you!

constellations/gemini-user/src/account/account.js
```diff
@@ -0,0 +1,95 @@
...
+require("./account.sass");
+import React from "react";
+import API from "../utils/api.js";
+import { union } from "lodash";
+
+import { SettingsTab } from "./settings.js";
+import {
+    RegistrationsTab,
+    RegistrationsTabMain,
+    RegistrationsTabAllClasses,
```
ahsu1230, line 11: I think these aren't used in this file

constellations/gemini-user/src/account/account.js
```diff
@@ -0,0 +1,95 @@
...
+                        key={index}>
+                        {item}
+                    </div>
+                );
+            }
+        );
+
+        let tabContent;
+        switch (this.state.selectedTab) {
+            case "settings":
```
ahsu1230, line 60: same use `TAB_SETTINGS:` instead of `"settings":`. Less prone to typos.

constellations/gemini-user/src/account/account.sass
```diff
@@ -0,0 +1,302 @@
...
+    input
+        padding: $spacing1x
+        border: 1px solid $colorGrayLight
+        border-radius: $spacing1x
+
+    a:hover
+        cursor: pointer
+
+    .red
+        color: #d91211
```
ahsu1230, line 292: can you use the color defined in `_constants.sass` (or define it yourself there? like `colorRed`)

constellations/gemini-user/src/account/payment.js
```diff
@@ -0,0 +1,109 @@
+"use strict";
+require("./account.sass");
```
ahsu1230, line 2: Similarly I think you should split your sass files. Keep each of these components isolated. Their JS files should be separate, and so should their CSS as well.

constellations/gemini-user/src/account/account.sass
```diff
@@ -0,0 +1,302 @@
...
+                display: flex
+                flex-direction: row
+                flex-wrap: wrap
+                border-bottom: 1px solid $colorWhiteLight
+                text-align: left
+
+                li
+                    padding: $spacing1x 0
+
+        .tab-content#payment-tab
```
ahsu1230, line 92: I would split the CSS to their own individual components.
Let account.sass handle CSS that handles anything all the tabs share. For example, the tab switching highlight behavior, etc. should be handled by account.sass. And let account/payment.sass handle everything SPECIFICALLY inside the payment tab content.

Even if the file is small, you can do something like this:

```
#view-account .tab-content#payment-tab
    ...
```

constellations/gemini-user/src/account/registrations.js
```diff
@@ -0,0 +1,278 @@
...
+} from "./accountUtils.js";
+
+export class RegistrationsTab extends React.Component {
+    state = {
+        userClasses: [],
+
+        upcomingAFHs: [],
+        locationsById: {},
+
+        viewAllEnrolledClasses: false,
```
ahsu1230, line 22: can remove the extra spaces.

constellations/gemini-user/src/account/settings.js
```diff
@@ -0,0 +1,268 @@
...
+                        Navigator, including all user and class information.
+                    </p>
+                    <a className="red">Request to Delete Account...</a>
+                </div>
+            </div>
+        );
+    };
+}
+
+class PasswordChange extends React.Component {
```
ahsu1230, line 126: I think this component can be its own file, especially since it's its own little section in Settings.

PR #216, opened 2020-08-11
email flow for admin site

DanielLiu14:


DanielLiu14:
@srenaxu @ahsu1230

constellations/gemini-admin/src/emails/accountInfo.js
```diff
@@ -0,0 +1,134 @@
...
+        if (e.target.checked) {
+            users.map((user, index) => {
+                if (user.id == userId) {
+                    var email = this.state.selectedUserEmails;
+                    email.push(user.email);
+                    this.setState({
+                        selectedUserEmails: email,
+                    });
+                }
+            });
```
ahsu1230, line 35: Maybe you can use find instead of map here
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

ahsu1230: Once you find the target email, then you can edit the list of emails.
Also rename the variable `email` to `emails` to convey a list of emails (instead of one)

constellations/gemini-admin/src/emails/accountInfo.js
```diff
@@ -0,0 +1,134 @@
...
+        } else {
+            users.map((user, index) => {
+                if (user.id == userId) {
+                    var email = this.state.selectedUserEmails;
+                    email.splice(email.indexOf(user.email), 1);
+                    this.setState({
+                        selectedUserEmails: email,
+                    });
+                }
+            });
```
ahsu1230, line 45: I think you may be able to use .filter() here.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

constellations/gemini-admin/src/emails/accountInfo.js
```diff
@@ -0,0 +1,134 @@
...
+        const id = this.props.id;
+        const email = this.props.email;
+        const users = this.props.users;
+        const transactions = this.props.transactions;
+        const name = this.props.name;
+
+        var emails = [];
+        users.map((user, index) => {
+            emails.push(user.email);
+        });
```
ahsu1230, line 59: one line: `emails = users.map((user) => user.email)`

constellations/gemini-admin/src/emails/accountInfo.js
```diff
@@ -0,0 +1,134 @@
...
+                        selectedUserEmails: email,
+                    });
+                }
+            });
+        }
+    };
+
+    render = () => {
+        const id = this.props.id;
+        const email = this.props.email;
```
ahsu1230, line 51: accountEmail, makes it less confusing which email this is referring to

constellations/gemini-admin/src/emails/accountInfo.js
```diff
@@ -0,0 +1,134 @@
...
+            status += user.email == email ? ", primary contact)" : ")";
+
+            if (this.state.selectedUserEmails.length > 0) {
+                this.state.selectedUserEmails.map((existingEmail, index) => {
+                    if (emails.indexOf(existingEmail) == -1) {
+                        this.setState({
+                            selectedUserEmails: [],
+                        });
+                    }
+                });
```
ahsu1230, line 72: filter or find here. Or maybe use forEach.

map() is used to create a list from another list.

constellations/gemini-admin/src/emails/accountInfo.js
```diff
@@ -0,0 +1,134 @@
...
+                    />
+                    <span className="column">{getFullName(user)}</span>
+                    <span className="column">{user.email}</span>
+                    <span className="column status">{status}</span>
+                </div>
+            );
+        });
+
+        var balance = 0;
+        transactions.map((transaction, index) => {
```
ahsu1230, line 88: this is a good place for `.forEach` instead of map. Map should be used when you want to create a new list from an original list. With forEach, you simply loop through all the elements and apply a function to each element.

At the end of the day, it's the same thing, but from a semantics perspective, it's more accurate.

constellations/gemini-admin/src/emails/accountInfo.js
```diff
@@ -0,0 +1,134 @@
...
+            <section id="account-info">
+                <span id="account-number">Account Information</span>
+                <h2>AccountId: {id}</h2>
+                <h2>Primary Email: {email}</h2>
+                <div id="account-users">
+                    <h2>Users in Account</h2>
+                    {userRows}
+                </div>
+                <div id="transaction-footer">
+                    <span>
```
ahsu1230, line 103: is this span needed?

constellations/gemini-admin/src/emails/accountInfo.js
```diff
@@ -0,0 +1,134 @@
...
+                        Account Balance: <b>{this.formatCurrency(balance)}</b>
+                    </span>
+                </div>
+                <div className="email-template">
+                    <span id="template-title">Generated Email Template</span>
```
ahsu1230, line 108: I think this is also supposed to be a header. For headers use `<hx>`. spans are usually for like little pieces of texts that are adjacent to each other. Like buttons or a horizontal list of keywords.

constellations/gemini-admin/src/emails/accountInfo.js
```diff
@@ -0,0 +1,134 @@
...
+                    <h3>To: {this.state.selectedUserEmails.toString()}</h3>
+                    <h3>
+                        Subject: Math Navigator: Account Balance Payment
+                        Reminder
+                    </h3>
+                    <h3>Message: </h3>
+                    <div className="generated-email">
+                        <p>Hello {name},</p>
```
ahsu1230, line 116: call this `primaryContactName` to ensure we're always talking to the person who's the primary contact / owner of the account.

constellations/gemini-admin/src/emails/accountInfo.js
```diff
@@ -0,0 +1,134 @@
...
+
+        var balance = 0;
+        transactions.map((transaction, index) => {
+            const amount = transaction.amount;
+            balance += parseInt(amount);
+        });
+
+        return (
+            <section id="account-info">
+                <span id="account-number">Account Information</span>
```
ahsu1230, line 95: this is supposed to be a header

constellations/gemini-admin/src/emails/accountInfo.js
```diff
@@ -0,0 +1,134 @@
...
+        });
+
+        var balance = 0;
+        transactions.map((transaction, index) => {
+            const amount = transaction.amount;
+            balance += parseInt(amount);
+        });
+
+        return (
+            <section id="account-info">
```
ahsu1230, line 94: This section isn't too large, so I think it would usually be okay to keep it as a single component. It will probably make your life a bit easier. But the other page will probably need to divide into multiple components, so if it's not too hard, I actually recommend splitting this component into two components as well to match code organization.

It would be nice when viewing this code and other EmailToProgram code, they have similar code organizations, so it should be easy to follow. For example, in this page, One component should be the AccountInformation, the other component is the GeneratedEmail section. In the other page, you'll probably have like three sections (ProgramInfo, SelectUserList, GeneratedEmail). Since there are so many components, you'll end up wanting to have them in different files.

So would probably even name your current folder to `emailPayment`. And when creating the other page, place all the stuff into the folder `emailProgram`.

constellations/gemini-admin/src/emails/email.js
```diff
@@ -0,0 +1,241 @@
...
+import { YesNoModal } from "../modals/yesnoModal.js";
+import { AccountInfo } from "./accountInfo.js";
+
+export class PaymentEmails extends React.Component {
+    state = {
+        id: getCurrentAccountId() || 0,
+        email: "",
+        users: [],
+        transactions: [],
+        selectedUserEmails: [],
```
ahsu1230, line 18: is this needed?

constellations/gemini-admin/src/emails/email.js
```diff
@@ -0,0 +1,241 @@
...
+        this.state.users.map((user, index) => {
+            if (index === 0) {
+                const name = user.firstName;
+                names += name;
+            }
+        });
+
+        return (
+            <div id="view-account">
+                {modalDiv}
```
ahsu1230, line 181: how would this get triggered? Is there a way to trigger this modal popping up?

constellations/gemini-admin/src/emails/email.js
```diff
@@ -0,0 +1,241 @@
...
+        });
+    };
+
+    renderModal = (showDeleteModal, onModalDeleteConfirm, onModalDismiss) => {
+        let modalDiv;
+        let modalContent;
+        if (showDeleteModal) {
+            modalContent = (
+                <YesNoModal
+                    text={"Are you sure you want to delete?"}
```
ahsu1230, line 110: are we deleting anything on this page?

constellations/gemini-admin/src/emails/email.js
```diff
@@ -0,0 +1,241 @@
...
+import React from "react";
+import axios from "axios";
+import { Link } from "react-router-dom";
+import API, { executeApiCalls } from "../api.js";
+import { getCurrentAccountId, setCurrentAccountId } from "../localStorage.js";
+import { Modal } from "../modals/modal.js";
+import { YesNoModal } from "../modals/yesnoModal.js";
+import { AccountInfo } from "./accountInfo.js";
+
+export class PaymentEmails extends React.Component {
```
ahsu1230, line 12: name this file emailPayments.js. There are going to be other email pages, so we want to differentiate them

constellations/gemini-admin/src/emailPayment/accountInfo.js
```diff
@@ -0,0 +1,151 @@
...
+    onCheckUser = (e, userId) => {
+        const users = this.props.users;
+        if (e.target.checked) {
+            var emails = this.state.selectedUserEmails;
+            var currentSelectedUsers = this.state.selectedUsers;
+            const checkedUser = users.find((user) => user.id == userId);
+            emails.push(checkedUser.email);
+            currentSelectedUsers.push(checkedUser);
+            this.setState({
+                selectedUserEmails: emails,
```
ahsu1230, line 29: I think let's get rid of this. We already keep track of the selected Users.
If you ever want only emails for selected Users, you can always use `selectedUsers.map(user => user.email)` which will create a list of emails from the list of users.

Having two variables of similar things and always having to update both can be messy and introduces potential problems like forgetting to update one of them.

constellations/gemini-admin/src/emailPayment/accountInfo.js
```diff
@@ -0,0 +1,151 @@
...
+                        ) == undefined
+                    ) {
+                        this.setState({
+                            selectedUserEmails: [],
+                            selectedUsers: [],
+                        });
+                        break;
+                    }
+                }
+            }
```
ahsu1230, line 80: can we move this into a function?
Like `const userRows = createUserRows(users, this.state.selectedUsers)`

constellations/gemini-admin/src/emailPayment/accountInfo.js
```diff
@@ -0,0 +1,151 @@
...
+            status += user.email == accountEmail ? ", primary contact)" : ")";
+
+            if (this.state.selectedUsers.length > 0) {
+                for (var i = 0; i < this.state.selectedUsers.length; i++) {
+                    if (
+                        users.find(
+                            (user) => user == this.state.selectedUsers[i]
+                        ) == undefined
+                    ) {
+                        this.setState({
```
ahsu1230, line 73: This can be a bit cleaner if:

```
var user = users.find(...)
if (!user) {
   ...
}
```

In Javascript, there's this idea of "falsy" values meaning 0, false, undefined, NaN, "". All of these values can work with the `!` operator which will make them "truthy".

So !(undefined) evaluates to true

constellations/gemini-admin/src/emailPayment/accountInfo.js
```diff
@@ -0,0 +1,151 @@
...
+                    <span className="column">{user.email}</span>
+                    <span className="column status">{status}</span>
+                </div>
+            );
+        });
+
+        var balance = 0;
+        transactions.map((transaction, index) => {
+            const amount = transaction.amount;
+            balance += parseInt(amount);
```
ahsu1230, line 98: are amounts really coming in as strings? and not integers?

constellations/gemini-admin/src/emailPayment/accountInfo.js
```diff
@@ -0,0 +1,151 @@
...
+        const id = this.props.id;
+        const accountEmail = this.props.email;
+        const users = this.props.users;
+        const transactions = this.props.transactions;
+        const name = this.props.name;
+
+        var emails = [];
+        emails = users.map((user) => user.email);
+
+        const userRows = users.map((user, index) => {
```
ahsu1230, line 62: can we move this to a function?
`const userRows = createUserRows(...)`

I think this is big enough to be modularized into its own function.

constellations/gemini-admin/src/emailPayment/accountInfo.js
```diff
@@ -0,0 +1,151 @@
...
+                        </span>
+                        <h3>To: {this.state.selectedUserEmails.toString()}</h3>
+                        <h3>
+                            Subject: Math Navigator: Account Balance Payment
+                            Reminder
+                        </h3>
+                        <h3>Message: </h3>
+                        <div className="generated-email">
+                            <p>Hello {name},</p>
+                            <p>
```
ahsu1230, line 132: Can you add a line here in the email that says your current account balance is $____
which should be the `balance` variable you calculated above.

PR #217, opened 2020-08-11
negative balance accounts

mythicat554:


constellations/orion/src/repos/account_repo_test.go
```diff
@@ -106,6 +106,71 @@ func Test_SelectAccountByPrimaryEmail(t *testing.T) {
...
+		AddRow(
+			2,
+			now,
+			now,
+			sql.NullTime{},
+			"test2@gmail.com",
+			"password2",
+			-200,
+		)
+	mock.ExpectPrepare("^SELECT (.*) FROM accounts (.+)").
```
mythicat554, line 136: I wasn't really sure how to mock the sql statement and match up the expressions

constellations/orion/src/controllers/account_controller_test.go
```diff
@@ -105,6 +105,45 @@ func TestGetAccounts_Failure(t *testing.T) {
...
+	var accountSums []domains.AccountSum
+	if err := json.Unmarshal(recorder.Body.Bytes(), &accountSums); err != nil {
+		t.Errorf("unexpected error: %v\n", err)
+	}
+	assert.EqualValues(t, 1, accountSums[0].Id)
+	assert.EqualValues(t, "test@gmail.com", accountSums[0].PrimaryEmail)
+	assert.EqualValues(t, "password", accountSums[0].Password)
+	assert.EqualValues(t, 2, accountSums[1].Id)
+	assert.EqualValues(t, "test2@gmail.com", accountSums[1].PrimaryEmail)
+	assert.EqualValues(t, "password2", accountSums[1].Password)
```
ahsu1230, line 144: Since this test relates to balance, can you also test the balances are correct?

constellations/orion/src/domains/account.go
```diff
@@ -18,6 +18,16 @@ type Account struct {
 	Password     string       `json:"password" db:"password"`
 }
 
+type AccountSum struct {
+	Id           uint         `json:"id"`
+	CreatedAt    time.Time    `json:"-" db:"created_at"`
+	UpdatedAt    time.Time    `json:"-" db:"updated_at"`
+	DeletedAt    sql.NullTime `json:"-" db:"deleted_at"`
+	PrimaryEmail string       `json:"primaryEmail" db:"primary_email"`
+	Password     string       `json:"password" db:"password"`
```
ahsu1230, line 27: Hmm so what I'm worried about here is whenever Account gets a new field, it would be wise to add that field here again, but it's easy to forget.

To make our life easier, have this struct only have two things:
```
type AccountSum struct {
    Account Account
    Balance int
}
```

constellations/orion/src/domains/account.go
```diff
@@ -18,6 +18,16 @@ type Account struct {
...
+	Sum          int          `json:"sum" db:"sum"`
```
ahsu1230, line 28: So we don't need the `db:"sum"` here because this sum never gets written into a database column. It's simply a field just for JSON responses, so we only need the json part.

constellations/orion/src/repos/account_repo.go
```diff
@@ -72,6 +73,38 @@ func (acc *accountRepo) SelectByPrimaryEmail(primary_email string) (domains.Acco
 	return account, errScan
 }
 
+func (acc *accountRepo) SelectAllNegativeBalances() ([]domains.AccountSum, error) {
+	results := make([]domains.AccountSum, 0)
+
+	statement := "SELECT accounts.*,SUM(amount) FROM accounts JOIN transactions ON accounts.id=transactions.account_id GROUP BY account_id HAVING SUM(amount) < 0"
```
ahsu1230, line 79: space between `accounts.*, SUM(amount)`

ahsu1230: but very nice :)

constellations/orion/src/router/handler.go
```diff
@@ -95,6 +95,7 @@ func (h *Handler) SetupApiEndpoints() {
 	{
 		apiAccounts.POST("/create", controllers.CreateAccount)
 		apiAccounts.GET("/account/:id", controllers.GetAccountById)
+		apiAccounts.GET("/unpaid", controllers.GetNegativeBalanceAccounts)
```
ahsu1230, line 98: can you order to have the GETs first then the POST after

constellations/orion/src/tests_integration/account_integration_test.go
```diff
@@ -170,3 +198,29 @@ func assertAccount(t *testing.T, id int, account domains.Account) {
 		assert.EqualValues(t, "password2", account.Password)
 	}
 }
+
+func createTransactions(t *testing.T) {
+	trans1 := createTransaction(1, 100, domains.PAY_PAYPAL, "notes1", 1)
+	trans2 := createTransaction(2, 200, domains.PAY_CASH, "notes2", 2)
+	trans3 := createTransaction(3, 300, domains.PAY_CHECK, "notes3", 3)
+	trans4 := createTransaction(4, -400, domains.CHARGE, "notes4", 1)
+	trans5 := createTransaction(5, -300, domains.CHARGE, "notes4", 2)
```
ahsu1230, line 207: notes5? You could also test with notes being empty string ""

constellations/orion/src/tests_integration/account_integration_test.go
```diff
@@ -75,6 +75,34 @@ func Test_GetAccountById(t *testing.T) {
...
+	}
+	assert.EqualValues(t, 1, accountSums[0].Id)
+	assert.EqualValues(t, "john_smith@example.com", accountSums[0].PrimaryEmail)
+	assert.EqualValues(t, "password1", accountSums[0].Password)
+	assert.EqualValues(t, -300, accountSums[0].Sum)
+	assert.EqualValues(t, 2, accountSums[1].Id)
+	assert.EqualValues(t, "bob_smith@example.com", accountSums[1].PrimaryEmail)
+	assert.EqualValues(t, "password2", accountSums[1].Password)
+	assert.EqualValues(t, -100, accountSums[1].Sum)
+
```
ahsu1230, line 101: Very nicely done testing. In addition, let's test `accountSums` is only size 2. There is NOT a third element in this array because account3 has non-negative account balance!

ahsu1230:
@srenaxu this is almost close to merging. When this does, you can use this endpoint to get unpaid accounts!

constellations/orion/src/repos/account_repo_test.go
tonywu315: This works, but you could also make it match up more closely if you want. You just need to make sure the regex matches `SELECT accounts.*,SUM(amount) FROM accounts JOIN transactions ON accounts.id=transactions.account_id GROUP BY account_id HAVING SUM(amount) < 0`.

PR #218, opened 2020-08-11
Fixed formatting and input fields

tonywu315:


constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -99,6 +99,10 @@ export class ClassEditPage extends React.Component {
                         sessions: sessions,
 
                         fullState: classObj.fullState,
+                        googleClassCode: classObj.googleClassCode || "",
+                        priceLump: classObj.priceLump || "",
+                        pricePerSession: classObj.pricePerSession || "",
```
ahsu1230, line 104: priceLump and pricePerSession are usually integers, so you can use: `classObj.priceLump || 0`, which will default to 0 if not available.

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -144,11 +152,9 @@ export class ClassEditPage extends React.Component {
             times: this.state.inputTimeString,
             fullState: this.state.fullState,
             googleClassCode: this.state.googleClassCode,
-            priceLump: this.state.priceLump,
-            pricePerSession: this.state.pricePerSession,
+            priceLump: parseInt(this.state.priceLump),
+            pricePerSession: parseInt(this.state.pricePerSession),
```
ahsu1230, line 156: and then we don't need this parseInt-ing

constellations/gemini-admin/src/classes/classEdit.js
```diff
@@ -206,8 +212,49 @@ export class ClassEditPage extends React.Component {
...
 
-    render = () => {
-        const title = this.state.isEdit ? "Edit Class" : "Add Class";
+    renderModal = (
+        showSaveModal,
+        showDeleteModal,
+        onModalOkSaved,
+        onModalDeleteConfirm,
+        onModalDismiss
+    ) => {
```
ahsu1230, line 221: I agree handling modals per page can be a bit messy, especially because every page can potentially have many diff modals. I wonder if we should remove all class-related modals to a diff file? Or even since modals across all pages are similar, wonder if we can move it to the utils folder?? (just lightly thinking about it).

For this case though, I think it would make more sense if this function took in a string "key" of some sort.
And inside this function is a switch statement. The reasoning for this is It's possible a bad developer could use this function like `renderModal(true, false, true, false)` which is dumb, but could happen. You want to make it easier on a developer by removing as much "potential" problems with function usage.
With the switch statement, the usage becomes more straight forward: `renderModal("askDelete")`, `renderModal("confirmDelete")`, `renderModal("confirmSave")`, `renderModal("dismiss")` or something like that.

constellations/orion/src/domains/class.go
```diff
@@ -48,10 +48,10 @@ func (class *Class) Validate() error {
 		return errors.New("invalid times")
 	}
 
-	//Price validation
-	//Both valid
-	if priceLump.Valid && pricePerSession.Valid {
-		return errors.New("invalid price: both valid")
+	// Price validation
+	// Both valid
+	if priceLump.Valid == pricePerSession.Valid {
```
ahsu1230, line 53: I see this was my fault. I originally told Chujia that a class COULD be free or something. But that almost never happens. If we have a "free" class, it's usually like a one-off seminar, which is outside the scope of what a "class" represents. Thanks for fixing this.

constellations/gemini-admin/src/classes/classEdit.js
ahsu1230: And you can create string constants at the top of the file so it's clear what is available for use.
```
MODAL_CONFIRM_DELETE = "confirm_delete";
MODAL_DISMISS = "dismiss";

...
renderModal(MODAL_CONFIRM_DELETE);
```

tonywu315: I'll put this in another PR.

PR #219, opened 2020-08-14
Account user transaction 

mythicat554:


constellations/orion/src/repos/account_repo.go
```diff
@@ -117,32 +118,82 @@ func (acc *accountRepo) SelectAllNegativeBalances() ([]domains.AccountSum, error
...
+	_, err = stmt.Exec(
 		now,
 		now,
 		account.PrimaryEmail,
 		account.Password,
 	)
 	if err != nil {
-		return appErrors.WrapDbExec(err, statement, account)
+		tx.Rollback()
+		return err
```
ahsu1230, line 149: We don't want to just return `err` anymore. We want to wrap it with some context which is what the below function does.
`appErrors.WrapDbExec(err, statement, account)`

We're going to wrap it with a `ERR_DB_EXEC` so we know it's an error that relates to DB Executing and we also pass in information of what we were trying to execute. Check out the appErrors package!

constellations/orion/src/repos/account_repo.go
```diff
@@ -117,32 +118,82 @@ func (acc *accountRepo) SelectAllNegativeBalances() ([]domains.AccountSum, error
 	return results, nil
 }
 
-func (acc *accountRepo) Insert(account domains.Account) error {
+func (acc *accountRepo) InsertWithUser(account domains.Account, user domains.User) error {
+	tx, err := acc.db.Begin()
+	if err != nil {
+		return err
+	}
 	utils.LogWithContext("accountRepo.Insert", logger.Fields{"account": account})
```
ahsu1230, line 126: Make sure to log first thing in every function. We want to call this log thing first, then start the transaction

constellations/orion/src/repos/account_repo.go
```diff
@@ -117,32 +118,82 @@ func (acc *accountRepo) SelectAllNegativeBalances() ([]domains.AccountSum, error
 	return results, nil
 }
 
-func (acc *accountRepo) Insert(account domains.Account) error {
+func (acc *accountRepo) InsertWithUser(account domains.Account, user domains.User) error {
+	tx, err := acc.db.Begin()
+	if err != nil {
+		return err
```
ahsu1230, line 124: `appErrors.WrapDbPrepare(err, statement)` - see my comment below about "wrapping errors"

constellations/orion/src/repos/account_repo.go
```diff
@@ -117,32 +118,82 @@ func (acc *accountRepo) SelectAllNegativeBalances() ([]domains.AccountSum, error
...
+		"is_guardian," +
+		"account_id," +
+		"notes," +
+		"school," +
+		"graduation_year" +
+		") VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)"
+	stmt2, err := acc.db.Prepare(statement2)
+	if err != nil {
+		tx.Rollback()
+		return err
```
ahsu1230, line 169: wrapp

constellations/orion/src/repos/account_repo.go
```diff
@@ -117,32 +118,82 @@ func (acc *accountRepo) SelectAllNegativeBalances() ([]domains.AccountSum, error
...
+		user.Phone,
+		user.IsGuardian,
+		user.AccountId,
+		user.Notes,
+		user.School,
+		user.GraduationYear,
+	)
+	if err != nil {
+		tx.Rollback()
+		return err
```
ahsu1230, line 189: return wrap

constellations/orion/src/repos/account_repo.go
```diff
@@ -117,32 +118,82 @@ func (acc *accountRepo) SelectAllNegativeBalances() ([]domains.AccountSum, error
...
+	}
+
+	err = tx.Commit()
+	if err != nil {
+		log.Fatal(err)
```
ahsu1230, line 194: same thing. We can return the err. However, if you look in the appErrors package, you'll see we can wrap this around a `WrapDbTxCommit.

If the transaction failed to commit for some reason, I think we also need to rollback.
so
```
if err != nil {
    tx.Rollback()
    return appErrors.WrapDbCommit(err, ...)
}
```

constellations/orion/src/repos/account_repo.go
```diff
@@ -117,32 +118,82 @@ func (acc *accountRepo) SelectAllNegativeBalances() ([]domains.AccountSum, error
...
 		now,
 		now,
 		account.PrimaryEmail,
 		account.Password,
 	)
 	if err != nil {
-		return appErrors.WrapDbExec(err, statement, account)
+		tx.Rollback()
+		return err
+	}
```
ahsu1230, line 150: add another if statement here.

We should also actually validate the Db Result. This part might be a bit confusing. As you might've noticed LOTS of things can go wrong with SQL / MySQL. 

statement.Exec() provides a "Result" struct or an error. The "Result" struct has information about how many rows were updated for example. Here, we want to check at least 1 row gets updated in the account table.

So I would do something like this:
```
result1, err := stmt.Exec(...)

if err != nil {
		tx.Rollback()
		return appErrors.WrapExec(err)
}
if err = appErrors.ValidateDbResult(result1, 1, "account was not inserted"); err != nil {
    tx.Rollback()
    return appErrors.WrapDbExec(err, statement, account)
}
```

this is an exception (you don't have to do this for stmt.Query(...)), but here two things can go wrong. The first if handles the `err` case. The second handles the sql.Result struct that might have inconsistent information. I do all of that for you in the appErrors package.

constellations/orion/src/repos/account_repo.go
```diff
@@ -117,32 +118,82 @@ func (acc *accountRepo) SelectAllNegativeBalances() ([]domains.AccountSum, error
...
+		user.IsGuardian,
+		user.AccountId,
+		user.Notes,
+		user.School,
+		user.GraduationYear,
+	)
+	if err != nil {
+		tx.Rollback()
+		return err
+	}
```
ahsu1230, line 190: Similar as above. You checked the err here which is good! We also want to check result2 from stmt2.Exec(...) by using appErrors.ValidateDbResult(...) so we'll need another if-clause.

constellations/orion/src/repos/account_repo.go
```diff
@@ -117,32 +118,82 @@ func (acc *accountRepo) SelectAllNegativeBalances() ([]domains.AccountSum, error
...
+
+	err = tx.Commit()
+	if err != nil {
+		log.Fatal(err)
 	}
-	return appErrors.ValidateDbResult(execResult, 1, "account was not inserted")
+	return appErrors.WrapDbExec(err, statement, account)
```
ahsu1230, line 196: and finally out of the million things that could go wrong, if we get to this point and still no errors, we can just return nil.

Lol sorry about all these if clauses. It's a lot here because we're doing two db operations inside a transaction so there potentially could be a lot wrong here. We are only doing things locally on our computer, so errors are much less likely to happen, but when we launch this live, it's not as simple. The orion webserver and the database could potentially be on their own individual machines crunching data somewhere on Amazon Web Services in one of their data warehouses.

One of the most common things that could happen (why we need all these error checks) is in the middle of our DB operation, orion could lose connectivity to DbMySQL - many reasons, internet sucks, power goes out for a moment, orion server crashes from bad code, or even god forbid, the New York AWS data warehouse gets flooded and all transactions need to route to a different backup database in Toronto or LA.

The operation here probably takes like around a second, but in that time (in computer time), many things could go wrong and we need a way to make sure all cases are handled. Otherwise, you'll end up persisting data when you shouldn't. It is a Go problem to see so many if / err clauses - I'm hoping our new style with appErrors doesn't make it too bad / verbose.

constellations/orion/src/repos/account_repo_test.go
```diff
@@ -176,14 +176,15 @@ func Test_SelectAllNegativeBalances(t *testing.T) {
 }
 
 //
-// Create
+// Create User and Account
 //
-func TestInsertAccount(t *testing.T) {
+func Test_InsertAccountAndUser(t *testing.T) {
```
ahsu1230, line 181: I removed all underscores from test names, so this can just be `TestInsertAccountAndUser`

constellations/orion/src/repos/account_repo_test.go
```diff
@@ -192,8 +193,26 @@ func TestInsertAccount(t *testing.T) {
...
+			2,
+			domains.NewNullString(""),
+			domains.NewNullString("schoolone"),
+			domains.NewNullUint(2004),
+		).WillReturnResult(result)
+	mock.ExpectCommit()
 	account := getAccount()
-	err := repo.Insert(account)
+	user := getUser()
+	err := repo.InsertWithUser(account, user)
```
ahsu1230, line 215: nicely done! everything above ^

constellations/orion/src/repos/account_repo_test.go
```diff
@@ -202,6 +221,7 @@ func TestInsertAccount(t *testing.T) {
 	if err := mock.ExpectationsWereMet(); err != nil {
 		t.Errorf("Unfulfilled expectations: %s", err)
 	}
+
```
ahsu1230, line 224: nit: extra space here?

constellations/orion/src/repos/migrations/000040_modify_email_accounts.down.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE accounts
+    DROP COLUMN primary_email;
```
ahsu1230, line 2: To drop a UNIQUE constraint, do something like this:
```
ALTER TABLE table_name DROP UNIQUE (column1, column2, . . . )
```

constellations/orion/src/repos/migrations/000041_modify_email_users.down.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE users
+    DROP COLUMN email;
```
ahsu1230, line 2: same here

constellations/orion/src/tests_integration/account_integration_test.go
```diff
@@ -72,13 +71,52 @@ func TestGetAccountById(t *testing.T) {
 	assert.EqualValues(t, "john_smith@example.com", account.PrimaryEmail)
 	assert.EqualValues(t, "password1", account.Password)
 
+	utils.ResetTable(t, domains.TABLE_USERS)
+	utils.ResetTable(t, domains.TABLE_ACCOUNTS)
+}
+
+// Test: Create 2 accounts with the same email
+func Test_CreateSameEmailAccountsFailure(t *testing.T) {
```
ahsu1230, line 79: Yes nice test!!! Also remove _ from test names please!

constellations/orion/src/tests_integration/account_integration_test.go
```diff
@@ -72,13 +71,52 @@ func TestGetAccountById(t *testing.T) {
...
+
+	recorder1 := utils.SendHttpRequest(t, http.MethodPost, "/api/accounts/create", body)
+	assert.EqualValues(t, http.StatusOK, recorder1.Code)
+
+	utils.ResetTable(t, domains.TABLE_USERS)
 	utils.ResetTable(t, domains.TABLE_ACCOUNTS)
 }
 
 // Test: Create 3 Accounts, GetNegativeBalances()
 func Test_GetNegativeBalanceAccounts(t *testing.T) {
```
ahsu1230, line 117: Looks like I forgot to remove _ from integration test names. Can you remove them in this file?
`TestGetNegativeBalanceAccounts`

constellations/orion/src/tests_integration/transactions_integration_test.go
```diff
@@ -55,10 +56,13 @@ func TestCreateTransactions(t *testing.T) {
 	assert.EqualValues(t, 1, transactions[1].AccountId)
 
 	utils.ResetTable(t, domains.TABLE_TRANSACTIONS)
+	utils.ResetTable(t, domains.TABLE_USERS)
+	utils.ResetTable(t, domains.TABLE_ACCOUNTS)
 }
 
 // Test: Create 1 Transaction, Update, Get By ID
-func TestUpdateTransaction(t *testing.T) {
+func Test_UpdateTransaction(t *testing.T) {
```
ahsu1230, line 64: Also remove _ from test names

constellations/orion/src/tests_integration/user_integration_test.go
```diff
@@ -125,19 +136,30 @@ func TestGetUsersByAccountId(t *testing.T) {
 
 }
 
+// Test: Create 2 Users with the same email
+func Test_CreateSameEmailUsersFailure(t *testing.T) {
```
ahsu1230, line 140: very nice! also same, remove _ from test names

constellations/orion/src/repos/account_repo.go
```diff
@@ -86,17 +86,18 @@ func (acc *accountRepo) SelectByPrimaryEmail(primaryEmail string) (domains.Accou
...
 	stmt, err := acc.db.Prepare(statement)
 	if err != nil {
-		return nil, err
+		return nil, appErrors.WrapDbPrepare(err, statement)
 	}
 	defer stmt.Close()
 	rows, err := stmt.Query()
 	if err != nil {
-		return nil, err
+		return nil, appErrors.WrapDbQuery(err, statement)
```
ahsu1230, line 100: oh nice thanks!

ahsu1230:
This is good! and ready, but the problem is if you merge this in now, account-creation will be broken for every one - since this isn't "backwards-compatible". So @tonywu315 will need to put a change on Github and when that PR is ready, we can merge both at the same time, so there won't be a "break" time for people.

constellations/orion/src/repos/account_repo.go
```diff
@@ -117,32 +118,97 @@ func (acc *accountRepo) SelectAllNegativeBalances() ([]domains.AccountSum, error
...
+		tx.Rollback()
+		return appErrors.WrapDbExec(err, statement, account)
+	}
+	if err = appErrors.ValidateDbResult(result1, 1, "account was not inserted"); err != nil {
+		tx.Rollback()
 		return appErrors.WrapDbExec(err, statement, account)
 	}
-	return appErrors.ValidateDbResult(execResult, 1, "account was not inserted")
+
+	lastId, err := result1.LastInsertId()
```
ahsu1230, line 157: a little more specific would be nice: `lastAccountId` or `newAccountId`

ahsu1230:
This is good to merge! Thanks :)

PR #221, opened 2020-08-15
Improved semester edit and removed redundant sorting

tonywu315:


constellations/gemini-admin/src/semester/semesterEdit.js
```diff
@@ -45,23 +32,25 @@ export class SemesterEditPage extends React.Component {
...
+    };
 
-    handleIdChange(event, value) {
-        this.setState({ [value]: event.target.value });
+    handleIdChange = (event, value) => {
+        const newValue = event.target.value;
+        this.setState({ [value]: newValue });
         var possibleTitle =
-            event.target.value.substring(5) +
+            newValue.substring(5, 6).toUpperCase() +
```
ahsu1230, line 45: Add a comment here of what you're expecting. i.e. you're assuming the value looks like: `2020_fall` or 2021_winter.

PR #223, opened 2020-08-16
Added class error page

tonywu315:


constellations/gemini-user/src/error/error.js
```diff
@@ -0,0 +1,26 @@
+"use strict";
+require("./error.sass");
+import React from "react";
+import { Link } from "react-router-dom";
+import srcBroken from "../../assets/compass_broken.png";
+
+export class ErrorPage extends React.Component {
```
ahsu1230, line 7: I would actually put this in the class folder since this is ONLY an error page related to classes. Would probably call this component ClassErrorPage

constellations/gemini-user/src/class/class.js
```diff
@@ -133,46 +133,49 @@ export class ClassPage extends React.Component {
...
-                </div>
+        if (_.isEmpty(classObj))
+            return <ErrorPage classId={this.props.classId} />;
+        else
+            return (
+                <div id="view-class">
+                    {this.getTitle(program, semester, classObj)}
+
+                    <div id="class-description">
+                        <h4 className="red">
```
ahsu1230, line 144: instead of calling the class red, indicate what the state means. So like `fullState` or `full`

PR #224, opened 2020-08-19
added email to programs page

DanielLiu14:


DanielLiu14:
not even close to done but ill just put what i have for now here

constellations/gemini-admin/src/emailProgram/emailPrograms.js
```diff
@@ -0,0 +1,230 @@
...
+        return (
+            <div id="view-program-emails">
+                <section id="Select-program">
+                    <h1>Generate Email to Program</h1>
+
+                    <h2>Select a Program</h2>
+                    <select
+                        value={this.state.selectProgramId}
+                        onChange={(e) =>
+                            this.handleProgramChange(e, "selectProgramId")
```
ahsu1230, line 173: we needed the second argument in other pages because we wanted one function to handle multiple properties (like programId, programTitle, description, etc.). But here, we have a function to only handle changing the programId.

So this can just be `this.handleProgramChange(e)`
And the user-selected programId becomes e.target.value;

constellations/gemini-admin/src/emailProgram/emailPrograms.js
```diff
@@ -0,0 +1,230 @@
...
+                })
+            )
+            .catch((err) => {
+                console.log("Error: api call failed. " + err.message);
+            });
+    };
+
+    handleProgramChange = (event, value) => {
+        var selectedProgram = this.state.programs.find(
+            (program) => program.programId === event.target.value
```
ahsu1230, line 51: event.target.value is used often here, so I would just save it to a variable first.
```
const programId = event.target.value;

... program.programId === programId
```

constellations/gemini-admin/src/emailProgram/emailPrograms.js
```diff
@@ -0,0 +1,230 @@
...
+        });
+    };
+
+    onCheckClass = (e, classId) => {
+        if (e.target.checked) {
+            API.get("api/user-classes/class/" + classId)
+                .then((response) => {
+                    const users = response.data;
+                    const userIds = users.map((user) => user.userId);
+                    userIds.map((userId) => {
```
ahsu1230, line 74: remember map() is used to create a new list.
Use .forEach() if you're pushing things to an existing list.

constellations/gemini-admin/src/emailProgram/emailPrograms.js
```diff
@@ -0,0 +1,230 @@
+"use strict";
+require("./emailPrograms.sass");
+import axios from "axios";
+import React from "react";
+import API, { executeApiCalls } from "../api.js";
+import { getFullName } from "../utils/userUtils.js";
+
+export class EmailProgramsPage extends React.Component {
```
ahsu1230, line 8: // Can you add TODO comments at the top here?
// I think for the user section, we need a "Select All" and "Deselect All" buttons since there can potentially be a couple of classes and MANY users.

constellations/gemini-admin/src/emailProgram/emailPrograms.sass
```diff
@@ -0,0 +1,61 @@
...
+        font-weight: bold
+        font-size: $fontSize3
+
+    h3
+        margin: $spacing3x 0
+        line-height: $lineHeight1
+        font-size: $fontSize1
+
+    .generated-email
+        width: 1000px
```
ahsu1230, line 57: for width please use % not hard-coding

PR #225, opened 2020-08-20
afh subjects endpoint

mythicat554:


constellations/orion/src/router/handler.go
```diff
@@ -100,6 +100,7 @@ func (h *Handler) SetupApiEndpoints() {
 		apiAFH.GET("/afh/:id", controllers.GetAFHById)
 		apiAFH.POST("/afh/:id", controllers.UpdateAFH)
 		apiAFH.DELETE("/afh/:id", controllers.DeleteAFH)
+		apiAFH.GET("/subjects", controllers.GetAllAFHSubjects)
```
ahsu1230, line 103: Can we re-order the endpoints here to have all the GETs together than POST/Deletes after?
so I would do all, subjects, getById, create, update, delete

PR #226, opened 2020-08-20
program states endpoint

mythicat554:


constellations/orion/src/router/handler.go
```diff
@@ -18,6 +18,7 @@ func (h *Handler) SetupApiEndpoints() {
 		apiPrograms.GET("/program/:programId", controllers.GetProgramById)
 		apiPrograms.POST("/program/:programId", controllers.UpdateProgram)
 		apiPrograms.DELETE("/program/:programId", controllers.DeleteProgram)
+		apiPrograms.GET("/states", controllers.GetAllProgramStates)
```
ahsu1230, line 21: not your fault, but can we re-order the endpoints here?
Let's put all the GETs together and then the POSTs/others

so I would do:
all, states, getProgramById, Create, Update, Delete

PR #231, opened 2020-08-23
enzyme tests for afh and help

DanielLiu14:


constellations/gemini-admin/src/help/help.sass
```diff
@@ -9,13 +9,13 @@
         font-size: $fontSize4
     
     p
-        width: $spacing128x
+        width: 60%
```
ahsu1230, line 12: actually try to see if this works: Instead of 60%
```
width: 100%
box-sizing: border-box
```

ahsu1230: ^ is preferred

PR #234, opened 2020-08-23
documentation for console debugging

DanielLiu14:


resources/frontend/console_debugging.md
```diff
@@ -0,0 +1,19 @@
+## Debugging in Browser
+For the front-end, a lot of debugging is done by using inspect element on a browser. One of the tools this allows you to use is the console.
+
+### How do you see the Console?
+In order to see the console, you first need to inspect the webpage you are on. To do this, right click anywher on the page and click inspect.
+
+![alt text][logo]
+
+[logo]: https://thomas.vanhoutte.be/miniblog/wp-content/uploads/chrome-right-click-inspect.png "Logo Title Text 2"
```
ahsu1230, line 9: can you include this image in resources/images and then use relative pathing to refer to it?
Sometimes (very rarely), an image could get taken down for whatever reason and then we won't be able to refer to it anymore. Better to always refer from an "internal" source.

PR #235, opened 2020-08-24
Sql documentation

mythicat554:


resources/backend/sql_advanced.md
```diff
@@ -1,9 +1,22 @@
...
+```mysql
+SELECT accounts.*, SUM(amount) FROM accounts JOIN transactions ON accounts.id=transactions.account_id GROUP BY account_id HAVING SUM(amount) < 0
+```
+As seen above, GROUP BY is commonly used with aggregate functions (e.g. SUM, MAX, MIN).
+
 ## What is HAVING?
+[SQL HAVING](https://www.w3schools.com/sql/sql_having.asp) is a statement used where the WHERE keyword cannot be used with aggregate functions (e.g. SUM, MAX, MIN). For example, in this codebase, HAVING is used to select accounts with negative balances:
+```mysql
+SELECT accounts.*, SUM(amount) FROM accounts JOIN transactions ON accounts.id=transactions.account_id GROUP BY account_id HAVING SUM(amount) < 0
+```
```
ahsu1230, line 15: Do you think you can include the results that you get? In other words, basically we expect x columns. The first (x-1) columns are the account columns and the last column is the SUM of blah balh

resources/backend/sql_advanced.md
```diff
@@ -1,9 +1,26 @@
...
+This statement will then return a group of rows where each account has a negative sum. The first couple columns describe the account, such as the account id, the primary email, and the password, and the last column is the sum. 
+
+<img src="../images/sql_ss.png">
```
ahsu1230, line 19: I think you can actually use this:
`ALTERNATIVE_TEXT`

like so:
`SQL_GROUP_BY_IMAGE`

PR #238, opened 2020-08-28
Program class tests

tonywu315:


constellations/gemini-user/src/class/classError.test.js
```diff
@@ -0,0 +1,20 @@
...
+describe("Class Error Page", () => {
+    const component = shallow(<ClassErrorPage />);
+
+    test("renders", () => {
+        expect(component.exists()).toBe(true);
+        expect(component.find("h1").text()).toBe("Page Not Found");
+        expect(component.find("Link").text()).toBe("View our Programs");
+    });
+
+    test("renders classId", () => {
```
ahsu1230, line 14: just to be more clear, `renders invalid classId`

constellations/gemini-user/src/class/class.test.js
```diff
@@ -0,0 +1,76 @@
...
+        expect(component.find("p").at(5).text()).toBe("Sat. 1:00pm - 3:00pm");
+        expect(component.find("p").at(6).text()).toBe("No sessions scheduled");
+
+        expect(component.find("h3").at(2).text()).toBe("Pricing");
+        expect(component.find("p").at(7).text()).toBe(
+            "Price per session: $50.00"
+        );
+
+        expect(component.find("Link").at(2).text()).toBe("Contact Us");
+        expect(component.find("Link").at(3).text()).toBe("< More Programs");
```
ahsu1230, line 74: Hmm I think this would be more readable and easier to find bugs if you grouped them by their elements.
Put all the Links, p, h3 together like this:
```
expect(component.find("Link").at(0).text()).toBe("Programs");
expect(component.find("Link").at(1).text()).toBe("Register");
expect(component.find("Link").at(2).text()).toBe("Contact Us");
expect(component.find("Link").at(3).text()).toBe("< More Programs");
```

I understand it doesn't follow the order in which things appear in the component, but we don't have to do that here in the test.

PR #242, opened 2020-08-30
Documentation and bug fixes

tonywu315:


resources/frontend/api_calls.md
```diff
@@ -0,0 +1,105 @@
...
+
+# Successive API calls
+
+In `api.js`, there is a `executeApiCalls()` function that allows us to execute a list of API calls. This function takes in 3 parameters, a list of api calls, a success callback, and a fail callback. We can import this using
+
+```javascript
+import API, { executeApiCalls } from "../api.js";
+```
+
+`executeApiCalls()` executes the API calls in order. If they all succeed, the success callback is executed. Otherwise, the fail callback is executed.
```
ahsu1230, line 80: can we get a simple example of how to use executeApiCalls?

PR #246, opened 2020-09-07
Setup ELK using Docker-Compose

ahsu1230:


ahsu1230:
Resources for ES + Kibana:
https://www.youtube.com/watch?v=6bXSfjwQVIc&t=761s&ab_channel=MelvinL

Now ES runs at http://localhost:9200/ and Kibana runs at http://localhost:5601/

*NOTES* ES in dev environment MUST run on a single-node.
Kibana won't work until ES is up and running so it should depend on ES.
Docker was having local memory problems, so had to temporarily stop other services and change Docker-Dashboard settings: Docker > Resources > Advanced
Memory: 4GB, Disk image -> 72GB

PR #247, opened 2020-09-08
Introduce pressly/goose for mysql db migrations

ahsu1230:


ahsu1230:
This library is really not going to work because it doesn't actually invoke migrations programmatically, they want it to be invoked via the command line as a binary, which makes it kind of difficult to work with Docker. There's a gap time between when the container is created and when db-mysql is ready.

When the container is started up, we include the goose binary created in the migrations folder. This binary has all the registered migration Up/Down files. And just waits...

When orion is able to start up and connect to db-mysql, we now have to invoke the CLI using an `exec` command. Preferably we don't want to proceed until the command is completely finished.
But this feels so ghetto.

PR #253, opened 2020-10-06
remove account from user site

ahsu1230:


ahsu1230:
Decided users will not be exposed to accounts-related stuff :(
So better to delete so unit tests aren't blocking and to reduce code maintenance.

